PR #4, opened 2020-03-27
Add tensorflow keras dependencies to readme

rickybas:


dvida:
Thanks! I need to branch out certain modules so not all packages need to be installed just to use the trajectory solver.

PR #10, opened 2021-04-25
Small bugs in orbit and trajectory routines

markmac99:
I encountered a few crashes in the orbit and trajectory code, typically when orbits were not closed, and a was thus calculated as < 0.  This crashed the code to calculate T, mean anomaly, eccentric anomaly and last-perihelion. 

To avoid terminating the correlator, i put in some checks on a, and set variables accordingly. In principle i think we should be calculating the hyperbolic equivalents of the anomalies, but my conic section maths is too rusty to remember them... As for T, I set this to 1e9, and i set last-perihelion to nan, since a hyperbolic object cannot have a last-perihelion. :)


dvida:
Thanks again Mark. I don't think we should ignore everything in share, perhaps just add the files that need to be ignored individually? At some point I'll be updating some of the tables there.
Also, if we're setting the mean anomaly and the rest to NaN, why not set the period to NaN as well?

markmac99:
Happy to take your guidance on the files in share. I excluded them because they're pulled down when the installation is done and so come up as constant untracked files:)  I'll amend .gitignore to explicitly exclude the four files. 

Regarding period, I think this is used in other code and and i am not sure what will happen if we try to take Sqrt(NaN^3). I can test it out though.  Leave it with me.

markmac99:
As discussed, individually excluded the files in share, and set T to np.nan for hyperbolic orbits.

PR #11, opened 2021-11-16
Added Luminous Efficiency Plots

lmcfadd6:
Call plotLumEff(self.frag_list) wherever is needed to plot results

### DIFF
--- a/wmpl/MetSim/GUI.py
+++ b/wmpl/MetSim/GUI.py
@@ -23,7 +23,9 @@ from PyQt5.uic import loadUi
 from wmpl.Formats.Met import loadMet
 from wmpl.MetSim.GUITools import MatplotlibPopupWindow
 from wmpl.MetSim.MetSimErosion import runSimulation, Constants
+from wmpl.MetSim.MetSimErosionCyTools import luminousEfficiency, massLossRK4, decelerationRK4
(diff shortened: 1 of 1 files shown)


dvida:
A few minor changes before I merge it:
- from wmpl.Trajectory.Trajectory import * - improving everything from a namespace is bad practice because it might lead to clashes between objects of the same name. Could you just import individual things you need, or better yet, do something like wmpl.Trajectory.Trajectory.<function_you_need>
- DENS_CONST = np.array([6.96795507e+01, ... - the actual air density coefficients for a given fireball are in self.const.dens_co. I recommend passing const into calcLumEff and using that variable instead of hardcoding these coeffs.
- docstring missing for plotLumEff
- calcLumEff("Leading", self.main_height_arr, self.main_mass_arr - the name is for the leading fragment, but the arrays for the main fragment are used?
-

dvida:
Thanks! Just one final thing - I don't see this used anywhere in the code, do we need it:
from wmpl.Trajectory.Trajectory import Trajectory, ObservedPoints, PlaneIntersection

Also, does it take long to compute the luminous efficiency? Can the loop perhaps never converge, and grind the normal operation of the simulation to a halt?

If you'd like, I can add a separate button for the lum eff plot to show up.

dvida:
This was implemented another way. Closing.

PR #17, opened 2022-06-28
Add OSM map trajectory plot

satmonkey:


markmac99:
I dropped this code into my own fork of WMPL and tested it yesterday but got an error at line 5174 of Trajectory.py. I think you will need to specify the additional modules that need to be installed (looks like cartopy?) and the additional optional features to support the map resolution.  
The error indicates i need to do "conda install -c conda-forge basemap-data-hires" or the equivalent for pip. but if i install basemap-data-hires, then WMPL silently fails - crashes in OSTools.loadBasemap but without any error message, 
I think installing highres data is incompatible with the basemap version that WMPL uses so i would be interested to know what platform you developed this on, and what exact versions of basemap basemap-data-hires and matplotlib you are using. I can then try with those exact versions. 

Full error attached. 
---------
OSError: Unable to open boundary dataset file. Only the 'crude' and  'low',
resolution datasets are installed by default.
------------

Traceback (most recent call last):
  File "C:\users\mark\anaconda3\envs\ukmon-shared\lib\site-packages\mpl_toolkits\basemap\__init__.py", line 1230, in _readboundarydata
    bdatfile = open(os.path.join(basemap_datadir,name+'_'+self.resolution+'.dat'),'rb')
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\users\\mark\\anaconda3\\envs\\ukmon-shared\\Library\\share\\basemap\\gshhs_i.dat'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\users\mark\anaconda3\envs\ukmon-shared\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\users\mark\anaconda3\envs\ukmon-shared\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "E:\DEV\meteorhunting\WesternMeteorPyLib\wmpl\Formats\ECSV.py", line 252, in <module>
    traj = solveTrajectoryGeneric(jdt_ref, meteor_list, dir_path, solver=cml_args.solver, \
  File "E:\DEV\meteorhunting\WesternMeteorPyLib\wmpl\Formats\GenericFunctions.py", line 291, in solveTrajectoryGeneric
    traj = traj.run()
  File "E:\DEV\meteorhunting\WesternMeteorPyLib\wmpl\Trajectory\Trajectory.py", line 6389, in run
    traj_best.savePlots(mc_output_dir, mc_file_name, show_plots=self.show_plots)
  File "E:\DEV\meteorhunting\WesternMeteorPyLib\wmpl\Trajectory\Trajectory.py", line 5174, in savePlots
    m = GroundMap(lat_list, lon_list, border_size=50, color_scheme='light')
  File "E:\DEV\meteorhunting\WesternMeteorPyLib\wmpl\Utils\PlotMap.py", line 115, in __init__
    self.m = Basemap(projection='gnom', lat_0=np.degrees(lat_mean), \
  File "C:\users\mark\anaconda3\envs\ukmon-shared\lib\site-packages\mpl_toolkits\basemap\__init__.py", line 1119, in __init__
    self._readboundarydata('gshhs',as_polygons=True)
  File "C:\users\mark\anaconda3\envs\ukmon-shared\lib\site-packages\mpl_toolkits\basemap\__init__.py", line 1233, in _readboundarydata
    raise IOError(msg)
OSError: Unable to open boundary dataset file. Only the 'crude' and  'low',
resolution datasets are installed by default.
If you are requesting an, 'intermediate', 'high' or 'full'
resolution dataset, you may need to download and install those
files separately with
`conda install -c conda-forge basemap-data-hires`.

satmonkey:
Hi Mark,
I only tested this on Ubuntu 18.04, virtual environment working in RMS.
And yes, some more lines are to be added to requirements:

**sudo apt-get install libproj-dev
pip install cartopy
pip uninstall shapely
pip install shapely --no-binary shapely**

can you run the above and try again,
Milan

markmac99:
Attached zip file with the ECSV files and the computed orbit pickle file. I procesed this with 
 python -m wmpl.Formats.ECSV . -l -x -r 20 -w

I wonder if the simplest fix is to make the map square? This would ensure adquate ground coverage. I am really keen on this addition by the way, i think it'll be very useful,.

satmonkey:
no really the square, but the 0 meridian was the issue.
Fixed by using Denis's approach by calculating the angular distance for setting the map extent.
Now it looks fine here on two different UK fireballs.

markmac99:
Seems to work :) However i suggest it would be cleaner if you just swapped lon_min and lon_max, in the case which crosses the meridian? Also you don't need the checks for lon_min < 0 and > 360 as Cartopy handles this internally and anyqway negative longitudes are valid data, :)

So my suggestion would be at line 130:
if lon_min > lon_max:
    lon_tmp = lon_min
    lon_min = lon_max
    lon_max = lon_tmp
    
Then remove "else" condition a few lines later which is not needed

markmac99:
ps - chart of our fireball produced with your most recent push - looks good ! 
20220511_233959_OSM_ground_track

HashtagPhysics:
In MATLAB, I use the built-in function wrapTo360 (or wrapTo180), to correct out-of range longitude values.  In Python, you can use the modulo operator %, to achieve the same, like this: angle % 360

Hope this helps! 

MATLAB CODE (the last line handles the special case of input = 360)
function lon = wrapTo360(lon)
%wrapTo360 Wrap angle in degrees to [0 360]
%
%   lonWrapped = wrapTo360(LON) wraps angles in LON, in degrees, to the
%   interval [0 360] such that zero maps to zero and 360 maps to 360.
%   (In general, positive multiples of 360 map to 360 and negative
%   multiples of 360 map to zero.)
%
%   See also wrapTo180, wrapToPi, wrapTo2Pi.

% Copyright 2007-2008 The MathWorks, Inc.

positiveInput = (lon > 0);
lon = mod(lon, 360);
lon((lon == 0) & positiveInput) = 360;

markmac99:
True, but its not necessary as Cartopy already handles wrapping of out of bounds longitudes. In fact, i suspect its converting longitude values > 180 into negative values (which is of course correct). 

By the way there's a simpler way to swap the values, without a temp variable
if min_lon > max_lon:
   min_lon, max_lon  = max_lon, min_lon

HashtagPhysics:
Do you assign an event id in your database?  I am curious the format, I have proposed a meteor event id format in my DRAFT document (Section 2), example as follows: 
image

I also propose (Section 3.8 - 3.9) extrapolating nominal lat/long to 40km to prevent duplicate ID's from multiple sources.  I haven't completed this yet in my code, but I think it is a good idea. 

[SCMED Database Spec Draft.docx](https://github.com/wmpg/WesternMeteorPyLib/files/9031077/SCMED.Database.Spec.Draft.docx)
:

markmac99:
Hi, Jim, the comments section of an unrelated github push are not the right place to ask questions about how the GMN refernces individual events. Can i suggest you join the globalmeteornetwork group on groups.io and ask there?  Also you could look at GMNs data which is available at https://globalmeteornetwork.org/data/traj_summary_data/. The data are publicly available and will probably answer your question.

markmac99:
unfortunately I've found another issue. Cartopy isn't available for the aarch64 architecture that i use on AWS, and a manual build  of cartopy on that hardware is erroring with No module named 'cartopy._version'. 

Can you please hold off integrating this into WMPL until i've found a solution ?

satmonkey:
maybe this would help?
https://kb.databricks.com/en_US/libraries/install-cartopy-on-cluster
We can comit to a new branch a merge into master once the  solution is found afterwards

markmac99:
The problem is more complex: there is a chain of dependencies and as aarch64 is a relatively new architecture only used for high end compute, most of the dependencies have to be built by hand! 
Cartopy 0.21 or later requires Python 3.8 or later
Cartopy 0.20 or later requires GEOS v 3.7.2
Geos 3.7.2 requires Proj 8.0.0 or later 
Proj requires sqlite3 v 3.37 or later. 

And once you build all those guys, cartopy itself won't build from source due to 
from ._version import version as __version__  # noqa: F401 
ModuleNotFoundError: No module named 'cartopy._version'

Ill figure it out but right now it is a showstopper!

markmac99:
ok, so i think we may need to be careful here because of the dependencies. The latest version of cartopy forces you to use at least Python 3.8, proj4 version 8.x or later, GEOS v3.7 or later and SQLlite3 v3.37 or later. These are generally available for consumer versions of Windows and Linux, but alll these versions are available for commercial versions of Linux (not available on Amazon Linux2 for example, which is a clone of Redhat Enterprise Linux, and uses the same repos).  I do now have it working on Amazon Linux 2 on the ARM64 Graviton platform, but it was quite complex and i had to rebuild all the libraries from source. 
Maybe, as long as we include suitable guidance and warnings, it will be ok.  Or maybe we need to make it optional somehow? So you can set a flag in setup.py that disables the cartopy dependency?

markmac99:
ok, this is getting super messy. Cartopy seems quite poorly supported except on Windows or a few consumer versions of Linux. I managed to build it on Graviton but i still can't get it working on x86-64 Fedora/RHEL clones, as the required versions of several packages are not available in the RHEL repos. I have tried building from source but i keep finding new dependencies not declared in the cartopy docs. Its really frustrating. Ultimately i need it to be able to run in a Docker container, which adds extra difficulties... I'll keep trying.

satmonkey:
on my Ubuntu, cartopy had to be built from source, version 0.18, using python 3.7.5. 
I think building from source will be essential, but pip can manage this automatically, if dependency problems.

markmac99:
in princple - but not in practice. I obviously did try that first :) "pip install cartopy" attempts to build every version of cartopy back to v 0.11 without success due to unmet dependencies. I think the problem is that you need specific versions of proj4, geos and sqlite which are not available for some server OSs. Server versions of popular OSes do not always contain the very latest versions of packages like this. RHEL for example is still on proj4 v 4.9 and GEOS v 3.4.2.  I'll find a solition but its frustrating ! :)

markmac99:
Unfortunately, so after several days working on this on various server OSes, i think the functionality has to be a separate utility, not automatically included in the main processing run.  Thats a shame as i really want it ! 

Unfortunately, cartopy is not easy to install on some platforms. 
If you have anaconda and access to conda-forge then it installs cleanly, but only if you install it *after* installing all other requied modules. However it forces a downgrade of some other modules used by WMPL. This could have unexpected side effects and reintroduces bugs and possible security flaws. If you try to install it at the same time as the other WMPL dependencies, the anaconda solver fails to find a valid solution because of the version conflicts. 

If one is unable to use anaconda, then there's no stable pip install for Windows, Debian, RHEL and derivants like Armbian and Amazon Linux2.  Again its dependent on versions of of several libraries not generally available on these Oses (same basic issue as with conda installs). 

I was able to build from source, but of course i also had to rebuild the dependencies sqlite3, geos and proj4, and it was pretty difficult to find all the right versions (which will evolve as cartopy is updated). That feels too technical for typical users, and if we do add this functionality, then then WMPL will need to include build scripts for windows and Linux that perform the necessary actions. 

I'm going to stop looking at this now as i have run out of time, but let me know if you find a solution to the issues, or make it a standalone module that can be called independently on demand, without forcing inclusion of cartopy in the main dataprocessing pipeline.

satmonkey:
Hi Mark,
I understand.
Let me then make this feature optional, by using the switch as a parameter. This should avoid the described obstacles.
I should be able to make the code modification in a few days, if my holiday time and girlfriend allow :)
Milan

satmonkey:
I have made the OSM map plot optional, via -o switch. Only in this case, the cartopy and related modules are loaded. Also, I noticed random error on Windows when loading OSM image tiles from the server, and fixed this by image spoofing (https://www.theurbanist.com.au/2021/03/plotting-openstreetmap-images-with-cartopy/). For some reason, I do see one of the commits twice, but it seems to be OK from the result view.

markmac99:
Great! I@ll take a look when i get a chance.

satmonkey:
After some improvements, it looks like this:
image

markmac99:
This is closed by #31  - i really must apologise to Milan because i solved all the issues about 6 months ago but completely failed to create an updated PR. The changes i made in PR31 ensure that it does not crash on platforms where Cartopy isn't available, and also works around a weird Windows bug where if cartopy is imported after basemap tehn loadBasemap() causes WMPL to silently crash. 

anyway this PR can be closed as #31 delivers the functionality

satmonkey:
Everything is fine, Mark. Will be happy to test it with the next fireball :)

PR #20, opened 2023-07-26
Improve OptimizePointingFOV util using parallel jobs

adalava:
This patch makes it use all available CPU by default when computing the best FOV overlap. The number of parallel jobs can be configured using parameter "parallel_jobs".

Tested using "time python -m wmpl.Utils.OptimizePointingsFOV" command on "Intel(R) Core(TM) i5-10300H CPU @ 2.50GHz", where computing time reduced from 15m25s to 3m29s. This is a single socket, 4 cores with 2 threads per socket, allowing the script to run 8 jobs at once.

### DIFF
--- a/wmpl/Utils/OptimizePointingsFOV.py
+++ b/wmpl/Utils/OptimizePointingsFOV.py
@@ -6,6 +6,7 @@ from __future__ import print_function, division, absolute_import
 
 import copy
 import datetime
+import multiprocessing
 
 import numpy as np
 import matplotlib.pyplot as plt
@@ -15,6 +16,7 @@ import scipy.optimize
 from wmpl.TrajSim.ShowerSim import initStationList, plotStationFovs
 from wmpl.Utils.TrajConversions import datetime2JD
 from wmpl.Utils.Math import estimateHullOverlapRatio
+from wmpl.Utils.PyDomainParallelizer import domainParallelizer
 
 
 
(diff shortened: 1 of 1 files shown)


markmac99:
Apart from that, it looks good! I don't use this functionality much as we have almost too many cameras in the UK (hence no optimal pointing direction for any new one!) but i know it is very useful and a performance boost would be good.

markmac99:
I saw the quick update - thanks ! All good from me.

adalava:
> You'll need to update the README to add joblib to the conda install instructions for people using conda.

Certainly! Patch updated, thanks!

dvida:
This is a great addition, thanks! We actually already have identical functionality that joblib has in wmpl, see wmpl.Utils.PyDomanParallelizer. This will allow us to drop the dependency on the new library and not break any existing installations. Would you mind changing the code? I think the interface is virtually identical and it should be an easy fix. Thanks!

adalava:
> This is a great addition, thanks! We actually already have identical functionality that joblib has in wmpl, see wmpl.Utils.PyDomanParallelizer. This will allow us to drop the dependency on the new library and not break any existing installations. Would you mind changing the code? I think the interface is virtually identical and it should be an easy fix. Thanks!

Sure! It was easy to use PyDomanParallelizer and the time measurements was exactly the same, thanks!

PR #27, opened 2024-11-14
Performance improvements in the trajectory solver

markmac99:
* Significant changes in CorrelateRMS.py and CorrelateEngine.py to support running the intersecting-planes and monte-carlo steps separately. A few specifics
* addition of new commandline parameters to set
  * operation mode (1= intersecting planes only, 2=mc only, 0/default = both modes)
  * repeat frequency in auto mode, default six hours, though one hour is actually ideal
  * set the maximum number of stations per solution, default 15 determined by testing various values
  * set the max number of trajectories to reload in monte-carlo mode, default 1000
* changes to the RMSDataHandle and JSON database classes to
  * refresh the database before each bucket is processed to ensure we're using the most recent data
  * persist the stage-1 trajectories immediately after being calculated
  * when in MC mode, reload the saved trajectories, run the MC phase to calc uncertainties and repersist the trajectories
  * update the code that loads the list of already-calculated trajectories
  * pick the best N stations based on some quality criteria
  * various cosmetic changes to add some logging and update python syntax

* Minor changes in Trajectory.py
  * add class member to the trajectory class to identify whether or not the MC phase has run
  * pass ignore_station flag through to various functions
  * (not related to this work) put try/except around the code that prints WGS84 data as these values are not available in older trajectories, and so it was causing a crash when reprocessing older data. 
  
* Cosmetic changes in AggregateAndPlot.py
  * provide more debug in the event of a solution being excluded from the reports. For example printing a message if the trajectory had too few points to be robust, a bad convergence angle etc. Previously these cases were silently skipped, but i felt it'd be useful to know why!  
  * update the Python syntax eg remove unnecessary \ between brackets, use is instead of == for boolean tests
  * add the noqa flag to a few more cases that my linter found objectionable due to excess whitespace

* small bugfixes in Math.py and TrajConversions.py to
  * handle a ValueError in jd2Date - i found a case where the jdt_ref was NaN.
  * add one extra bucket when getting the buckets to divide up data. Previously, data between 12:00UT and midnight was sometimes being missed and therefore delayed by 12-24 hours. 
  
I also removed the asteroid and comet data files and the numpy GMN shower table from the repo, and added them to .gitignore,  as these are downloaded afresh and compiled dynamically by setup,py and therefore should not be in github i think.


markmac99:
I am currently testing this solution on the UWO test infrastructure on a 16-core server as follows:
* Python 3.11 environment with numpy 1.26.x installed - be sure not to allow numpy v2 to install
* three instances of RMSCorrelate are running as follows:
  * one to run the initial non-MC phase to generate initial trajectories:
``` 
python -m wmpl.Trajectory.CorrelateRMS $datadir --cpucores 4 --logdir ./logs --mcmode 1 --outdir ./ --dbdir ./ --auto 6 --autofreq 60
```
  * two instances to run the monte-carlo simulations, These are timed so that the first one runs 15 minutes after the phase1 process and then to run 30 minutes apart: 
```
python -m wmpl.Trajectory.CorrelateRMS $datadir --cpucores 6 --logdir ./logs --mcmode 2 --outdir ./ --dbdir ./ --auto --autofreq 60
```
$datadir is /srv/meteor/rms/gmn/extracted_data.

markmac99:
Example script for phase 1. 
```
#!/bin/bash
source ~/.bashrc
eval "$(conda shell.bash hook)" > /dev/null
source ~/miniconda3/bin/activate wmpl311
export PYTHONPATH=$HOME/src/WesternMeteorPyLib
datadir=/srv/meteor/rms/gmn/extracted_data/
cd $HOME/data
python -m wmpl.Trajectory.CorrelateRMS $datadir --cpucores 4 --logdir ./logs --mcmode 1 --outdir ./ --dbdir ./ --auto 6 --autofreq 60
```

markmac99:
Example script for phase 2:
```
#!/bin/bash
source ~/.bashrc
eval "$(conda shell.bash hook)" > /dev/null
source ~/miniconda3/bin/activate wmpl311
export PYTHONPATH=$HOME/src/WesternMeteorPyLib
datadir=/srv/meteor/rms/gmn/extracted_data/
cd $HOME/data
python -m wmpl.Trajectory.CorrelateRMS $datadir --cpucores 8 --logdir ./logs --mcmode 2 --outdir ./ --dbdir ./ --auto --autofreq 60
```

markmac99:
Testing report attached.
https://d.docs.live.net/FA5EDB7AAC3B1A8A/Documents/Work/gmn/GMN%20Trajectory%20Solver%20Optimisation%20-%20update.docx
Let me know if you cant access it.

markmac99:
Some interesting differences in the output. The new solver sometimes finds solutions that the old solver missed or finds a different combination of stations to the new one. For example, the "old" solver attempted a 4-station solution of data from ES0001, ES000F, ES000H and ES000Q which failed, and no trajectory was created. However, the "new" solver found two viable solutions of two stations from the same data. 
I think this is caused by how frequently the solver runs, and how often data are available. The old solver processed these data at 16:43 on 11/11, at which time all four stations had uploaded. It then attempted a 4-station match which failed. 
The new solver processed ES0001 and ES000H at 08:32 to form one trajectory, and then received ES000F and ES000Q at 09:49 to form a second. It did attempt to improve the earlier solution but of course it failed for the same reason the four-station solution failed and so the "unused" observations were paired in a later phase. 
The two solutions in the new dataset have different orbital and trajectory characteristics which are outside each other's error bars, for example end heights 89.18+/-0.01 vs 92.02+/-0.03 and Vg 53.1+/-0.3  vs 53.9+/-0.3 but they're pretty close.

PR #39, opened 2025-03-06
Update SampleTrajectoryPositions.py

markmac99:
Bugfix to prevent a crash when the requested height range is outside the observed height range of the meteor. Issue reported by Milan.

### DIFF
--- a/wmpl/Utils/SampleTrajectoryPositions.py
+++ b/wmpl/Utils/SampleTrajectoryPositions.py
@@ -201,6 +201,7 @@ def sampleTrajectory(traj, beg_ht, end_ht, sample_step, show_plots=False):
     # Go through every distance from the Earth centre and compute the geo coordinates at the given distance,
     #   as well as the point-to-point azimuth and elevation
     prev_eci = None
+    good_data = False
     for ht, radius in zip(height_array, radius_array):
 
         # If the height is lower than the eng height, use a fixed velocity of 3 km/s
@@ -218,7 +219,12 @@ def sampleTrajectory(traj, beg_ht, end_ht, sample_step, show_plots=False):
         # Compute the intersection between the trajectory line and the sphere of radius at the given height
         intersections = lineAndSphereIntersections(np.array([0, 0, 0]), radius, traj.state_vect_mini, 
             traj.radiant_eci_mini)
-
+        
+        # if there are no intersections it means that the height was outside the observed start/end height of the meteor
+        if len(intersections) == 0:
+            continue
+        # if we get here we have at least one valid row of data 
+        good_data = True
 
         # Choose the intersection that is closer to the state vector
         inter_min_dist_indx = np.argmin([vectMag(inter - traj.state_vect_mini) for inter in intersections])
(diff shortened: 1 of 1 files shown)


markmac99 (commented):
I think actually the issue might be that Milan's particular trajectory could not be projected below a certain height of around 63 km, but the fix is still valid because it prevents a crash when intersections is an empty list, and therefore np.argmin() would crash. 
The intersections list is empty if there are no points on the trajectory which intersect with a sphere of diameter Re +h.  I suspect this can happen if the state vector is tangential to the Earth's surface and therefore has no projected points below a certain altitude. Either way the fix avoids a crash.

markmac99:
ps - without the fix you will get this, indicating that intersections was an empty list for h=60 for this particular dataset. 
```
(wmpl) PS E:\dev\meteorhunting\testing\wmpl_tests\milan_sampltrajpos> python -m wmpl.Utils.SampleTrajectoryPositions .\20250303_231256_trajectory.pickle 60 24 0.1
Beginning coordinates (observed):
    Lat: 52.242460
    Lon: 13.656528
    Elev: 66909.3

Ground-fixed azimuth and altitude:
 Time(s), Sample ht (m),  Lat (deg),   Lon (deg), Height (m), Azim (deg), Elev (deg)
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "e:\dev\meteorhunting\westernmeteorpylib\wmpl\Utils\SampleTrajectoryPositions.py", line 374, in <module>
    sampleTrajectory(traj, beg_ht, end_ht, sample_step, show_plots=True)
  File "e:\dev\meteorhunting\westernmeteorpylib\wmpl\Utils\SampleTrajectoryPositions.py", line 224, in sampleTrajectory
    inter_min_dist_indx = np.argmin([vectMag(inter - traj.state_vect_mini) for inter in intersections])
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\users\mark\miniconda3\envs\wmpl\Lib\site-packages\numpy\core\fromnumeric.py", line 1325, in argmin
    return _wrapfunc(a, 'argmin', axis=axis, out=out, **kwds)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\users\mark\miniconda3\envs\wmpl\Lib\site-packages\numpy\core\fromnumeric.py", line 56, in _wrapfunc
    return _wrapit(obj, method, *args, **kwds)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\users\mark\miniconda3\envs\wmpl\Lib\site-packages\numpy\core\fromnumeric.py", line 45, in _wrapit
    result = getattr(asarray(obj), method)(*args, **kwds)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: attempt to get argmin of an empty sequence
```

satmonkey:
I do confirm, this is working as expected now, thank you Mark!

dvida:
I usually set the input parameters to -1 -1 0.1 (sample from start to end every 100 m) so this issue never comes up.
Thanks for making explicit checks :)

PR #61, opened 2026-02-14
Enabling Distributed Processing

markmac99:
An upgrade to the RMS solver, CorrelateRMS, to enable distributed processing across multiple servers. To assist with this the first step has been split into two, one to create candidates and one to perform initial simple solutions.  

This PR also replaces the JSON database with SQLite which is necessary to enable distributed processing and also brings performance benefits. The PR adds two new commandline arguments, one to control how much data to retain in the databases, and one to post-fix the log name with the phase name eg correlate_rms_20260101_12345_cands.log. This is to ensure each phase's logfile is uniquely named and to make it easier to monitor and debug.


markmac99:
> Mark, just a question, is the rebound import needed for wmpl.Utils.Dcriteria?

I didn't actually add the Rebound functionality.  The "rebound" functionality was added by @tamidodo and Denis last year.

However, i just checked and Dcriteria.py has no dependency on rebound, so i am not sure how the error is arising for you.
i am able to import calcDSH without any issues, 
```
>>> from wmpl.Utils.Dcriteria import calcDSH
>>>
```
One possibility is that at some point in the past you "installed" the library by running python setup.py install. This would have put a precompiled copy in the lib/python3.10/site-packagesWesternMeteorPyLib directory in your virtualenv, and this is likely to be out of date.  There might then be dependencies which were since removed. Python will prefer to read from any copy in site-packages before reading files in the PYTHONPATH or local tree, so this can definirely cause issues. 

I now routinely delete the copy in site-packages to avoid issues. WMPL runs just fine "not installed", once you've obtained the required datafiles which will be in the share folder.  I keep meaning to write a standalone function to collect the files (orbit, shower, and ephemerides) without doing the whole install. 

ps I also think that the REBOUND module should be renamed to avoid the clash with the pypi module and Linux library of the same name.

dvida:
Brilliant work! I'm excited to try it. Here is what I run with Codex CLI found, could you possibly take a peek a there and see if they have any merit?

-----

I reviewed this PR in detail and found several issues that should be fixed before merge. The biggest risks are in the new remote/distributed processing paths, where transfer failures can currently be treated as success.

## 1. Remote upload retry helper returns success after all retries fail

`wmpl/Utils/remoteDataHandling.py`, `RemoteDataHandler.putWithRetry()`

Current code:

```python
def putWithRetry(self, local_name, remname):
    for i in range(10):
        try:
            self.sftp_client.put(local_name, remname)
            break
        except Exception:
            time.sleep(1)
    if i == 10:
        log.warning(f'upload of {local_name} failed after 10 retries')
        return False
    return True
```

Because `range(10)` ends with `i == 9`, `i == 10` is never true. If all ten upload attempts fail, this function still returns `True`.

That is dangerous because callers then move or delete local results as if upload succeeded, for example phase1 files and trajectory folders in `uploadToMaster()`.

Suggested fix:

```python
def putWithRetry(self, local_name, remname):
    for attempt in range(10):
        try:
            self.sftp_client.put(local_name, remname)
            return True
        except Exception as e:
            log.warning(f'upload attempt {attempt + 1}/10 failed for {local_name}: {e}')
            time.sleep(1)

    log.warning(f'upload of {local_name} failed after 10 retries')
    return False
```

## 2. Remote download can fail and still mark/remove the master copy

`wmpl/Utils/remoteDataHandling.py`, `collectRemoteData()`

Current code retries `self.sftp_client.get()`, but if all retries fail it still falls through to:

```python
self.sftp_client.rename(fullname, f'{rem_dir}/processed/{trajfile}')
```

or even:

```python
self.sftp_client.remove(fullname)
```

This can mark data as processed or delete it from the master even though the child did not successfully download it.

Suggested fix:

```python
downloaded = False

for attempt in range(10):
    try:
        self.sftp_client.get(fullname, localname)
        downloaded = True
        break
    except Exception as e:
        log.warning(f'download attempt {attempt + 1}/10 failed for {fullname}: {e}')
        time.sleep(1)

if not downloaded:
    log.warning(f'download of {fullname} failed after 10 retries; leaving remote file in place')
    continue

try:
    self.sftp_client.rename(fullname, f'{rem_dir}/processed/{trajfile}')
except Exception:
    log.warning(f'unable to move {fullname} to processed after successful download')
```

I would avoid removing the remote file as a fallback unless there is a strong reason; keeping it is safer than losing work.

## 3. Child phase-2 mode crashes due to undefined `mcmode`

`wmpl/Trajectory/CorrelateRMS.py`, `RMSDataHandle.getRemoteData()`

Current code:

```python
if self.mc_mode == MCMODE_PHASE1 or self.mc_mode == MCMODE_BOTH:
    status = self.RemoteDatahandler.collectRemoteData('candidates', self.output_dir, verbose=verbose)
elif mcmode == MCMODE_PHASE2:
    status = self.RemoteDatahandler.collectRemoteData('phase1', self.output_dir, verbose=verbose)
```

`mcmode` is not defined in this method. This will raise `NameError` when a child node starts in mode 2.

Suggested fix:

```python
if self.mc_mode == MCMODE_PHASE1 or self.mc_mode == MCMODE_BOTH:
    status = self.RemoteDatahandler.collectRemoteData('candidates', self.output_dir, verbose=verbose)
elif self.mc_mode == MCMODE_PHASE2:
    status = self.RemoteDatahandler.collectRemoteData('phase1', self.output_dir, verbose=verbose)
else:
    status = False
```

## 4. Phase1-only child upload can crash when no `trajectories` directory exists

`wmpl/Utils/remoteDataHandling.py`, `uploadToMaster()`

`traj_dir` is assigned only inside this block:

```python
if os.path.isdir(os.path.join(source_dir, 'trajectories')):
    traj_dir = f'{source_dir}/trajectories'
    ...
```

but later it is used unconditionally:

```python
if success_flag:
    shutil.rmtree(traj_dir, ignore_errors=True)
```

A child that only produced phase1 files and databases can crash with `UnboundLocalError`.

Suggested fix:

```python
traj_dir = os.path.join(source_dir, 'trajectories')

if os.path.isdir(traj_dir):
    for dirpath, dirnames, filenames in os.walk(traj_dir):
        ...
    if success_flag:
        shutil.rmtree(traj_dir, ignore_errors=True)
```

or at minimum:

```python
if success_flag and os.path.isdir(os.path.join(source_dir, 'trajectories')):
    shutil.rmtree(os.path.join(source_dir, 'trajectories'), ignore_errors=True)
```

## 5. AutoGUI no longer handles solver failure gracefully

`wmpl/Trajectory/AutoGUI.py`

This PR removes the previous `try/except` and `None` guard around `solveTrajectoryGeneric()`. If the solver raises or returns `None`, the GUI now crashes or dereferences `self.traj.output_dir`.

Suggested fix: restore the previous guard.

```python
try:
    self.traj = solveTrajectoryGeneric(
        jdt_ref, meteor_list, self.dir_path, solver=kwargs.solver,
        max_toffset=max_toffset, monte_carlo=False, save_results=False,
        geometric_uncert=kwargs.uncertgeom,
        gravity_correction=(not kwargs.disablegravity),
        gravity_factor=kwargs.gfact,
        plot_all_spatial_residuals=False,
        plot_file_type=kwargs.imgformat,
        show_plots=False,
        v_init_part=kwargs.velpart,
        v_init_ht=vinitht,
        show_jacchia=kwargs.jacchia,
        estimate_timing_vel=(False if kwargs.notimefit is None else kwargs.notimefit),
        fixed_times=kwargs.fixedtimes,
        mc_noise_std=kwargs.mcstd
    )
except Exception as e:
    print(f"Trajectory solution raised exception: {e}")
    self.worker.update_status.emit('red', f"Trajectory solution failed: {e}")
    return False

if self.traj is None:
    print("Trajectory solution failed (solver returned None).")
    self.worker.update_status.emit('red', "Trajectory solution failed!")
    return False
```

## 6. `CandidateDatabase.getCandidate()` reads the wrong column

`wmpl/Trajectory/CorrelateDB.py`

The candidates table is:

```python
CREATE TABLE candidates(cand_id VARCHAR UNIQUE, ref_dt REAL, obs_ids VARCHAR, status INTEGER)
```

but `getCandidate()` does:

```python
obs_ids = json.loads(rw[1])
```

`rw[1]` is `ref_dt`, not `obs_ids`.

Suggested fix:

```python
obs_ids = json.loads(rw[2])
```

or better, select only the needed column:

```python
cur = self.dbhandle.execute(
    "SELECT obs_ids FROM candidates WHERE cand_id=? and status=1",
    (cand_id,)
)
rw = cur.fetchone()
if rw is not None:
    obs_ids = json.loads(rw[0])
```

## Additional recommendation

Several SQLite queries are built with f-strings. Most inputs are internally generated, but using parameterized queries would make this much safer and less brittle, especially for observation IDs, candidate IDs, and file paths.

Example:

```python
cur = self.dbhandle.execute(
    "SELECT obs_id FROM paired_obs WHERE obs_id=? and status=1",
    (obs_id,)
)
```

## Minimal tests I would add

- `putWithRetry()` returns `False` after repeated SFTP failures and does not move/delete local files.
- `collectRemoteData()` leaves remote files in place when download fails.
- child mode 2 initialization calls `collectRemoteData('phase1', ...)` without raising `NameError`.
- `uploadToMaster()` succeeds when only `phase1/` and database files exist, with no `trajectories/` directory.
- AutoGUI handles `solveTrajectoryGeneric()` returning `None`.
- `CandidateDatabase.getCandidate()` returns the saved observation IDs.
```

**Comment by markmac99** [comment]
Thanks Denis,will take a fuller look at a couple of them but

1 is a bug, should be if i==9. Will fix it.

2 If we've failed ten times to download a file, then its extremely unlikely we'll be able to rename or delete it. I concede its possible, but ten failures in a row probably indicates some serious network issues. I'll review options.

3 is indeed a bug, though i suspect the condition is unreachable. Will fix it.

4 Saving a phase 1 solution creates both the phase1 version AND the trajectories folder tree and so the trajectories folder must exist if there's a phase1 solution. So this situation could only arise if a set of candidates resulted in zero solutions. This is possible of course though unlikely. I think the underlying problem here is that lines 314-319 should be intented one more tab.

5 - this is interesting as i have not touched this code. I'm guessing some change was made post my forking the master branch. I'll check it out and integrate any subsequent changes.

6 - this function is actually unused, so although the bug is real, it has no impact! i'll fix it or remove the unused function.

**Comment by markmac99** [comment]
Done some work to close points 1,2,3, and 4, and I've fixed the unused function mentioned in 6 above.  I'll run some tests today and advise. 

Issue 5 is a false alarm as this PR doesn't introduce any changes to `Trajectory/AutoGUI.py`. 

I can see however that a change was made a week ago to this file to add the try/except code mentioned in the Codex comment. These changes should not be overwritten by this PR.

**Comment by dvida** [comment]
Agreed that issue 5 is a false alarm, I fixed this recently. Not sure why it picked it up. Let me know once things are ready to rock and roll.

**Comment by markmac99** [comment]
Ha. During testing i have remembered why i was removing the remote copy even if the download failed.... 

When a child node downloads a file from the server, it then attempts to move it to the 'processed' folder on the server. This is to avoid reprocessing it again on the next pass. If the downloaded file can't be moved to the processed folder it is because a copy ALREADY exists in the "processed" folder, ie it was already processed. So, it should indeed be deleted from the server as otherwise the next pass will again download and process it, and so on ad infinitum. 

On the other hand if i can't download the file in the first place. its virtually certain to be because of a network glitch. If after 10s of retries it still can't be downloaded, then we do probably want to leave it, though the rename / delete would also almost certainly fail. 

I'll update the code to remove the file from the server if the move/rename operation fails.

**Comment by markmac99** [comment]
okay, all done and tested. During testing i spotted and corrected a couple of other small issues, and improved logging so its possible to more easily track which trajectories got assigned to child nodes.

**Comment by markmac99** [comment]
I think this is all good to go, pace the changes in documentation you asked for however i found one interesting problem.

When a monte-carlo solution is run on a phase1 solution, *sometimes* the reference timestamp changes by a few milliseconds. 

When run on a single machine, the software can detect this because each improved solution has the original trajectory path stored in the `pre_mc_longname` field and can therefore remove the old folder.  

However, if phase2 solutions are being distributed to child nodes, each node has its own copy of the files. The software will correctly delete the "old" copy on the child node, but when the data are uploaded to the server, the server simply moves the uploaded files to the trajectories folder. It does not check if the solution is an improvement to an existing solution and so sometimes, we can end up with two solutions on disk.  

Logic already exists to delete these when doing the MC phase on the master node, so i just need to extend this to run when merging in data from child nodes. 

If we prefer to wait till its resolved i am fine with that.  Should only take a few days. 

Meanwhile the workaround is simple - don't distribute the MCPhase  to child nodes  :)

**Comment by markmac99** [comment]
I've merged in all the recent upstream changes made via other PRs, and deployed this branch to the UKMON servers. No issues so far.

File: wmpl/Trajectory/CorrelateDB.py [hunk 1]
```diff
@@ -0,0 +1,1092 @@
...
+
+        table_name = 'failed_trajectories' if failed else 'trajectories'
+        if verbose:
+            log.info(f'getting trajectories between {jd2Date(jdt_start, dt_obj=True).strftime("%Y%m%d_%M%M%S.%f")} and {jd2Date(jdt_end, dt_obj=True).strftime("%Y%m%d_%M%M%S.%f")}')
+
+        if not jdt_end:
+            self.dbhandle.execute(f"SELECT * FROM {table_name} WHERE jdt_ref={jdt_start}")
+            rows = cur.fetchall()
+        else:
+            rows = self.dbhandle.execute(f"SELECT * FROM {table_name} WHERE jdt_ref>={jdt_start} and jdt_ref<={jdt_end}")
```
dvida: **Bug: this can return deleted trajectories, and the `not jdt_end` branch crashes.**

Two problems here:
1. Neither query filters on `status=1`, so trajectories that were marked removed (status set to 0) are still returned. `getComputedTrajectories()` in CorrelateRMS then treats them as live.
2. In the `if not jdt_end:` branch the cursor is never assigned to a variable, then `rows = cur.fetchall()` uses an undefined `cur` → `NameError`. (Also the following loop does `rows.fetchall()`, so both branches need to return a cursor, not a list.)

This fixes both:

```suggestion
        if not jdt_end:
            rows = self.dbhandle.execute(f"SELECT * FROM {table_name} WHERE jdt_ref={jdt_start} and status=1")
        else:
            rows = self.dbhandle.execute(f"SELECT * FROM {table_name} WHERE jdt_ref>={jdt_start} and jdt_ref<={jdt_end} and status=1")
```

wmpl/Trajectory/CorrelateDB.py
```diff
@@ -0,0 +1,1092 @@
...
+                        f"'{json.dumps(ign_obs_ids)}',1)")
+
+        sql_str = sql_str.replace('nan','"NaN"')
+        try:
+            self.dbhandle.execute(sql_str)
+        except Exception as e:
+            print(e)
+            print(sql_str)
+        self.dbhandle.commit()
+        return True
```
dvida: **Bug: a failed insert is hidden and the trajectory is reported as saved.**

On an exception this prints to stdout (invisible in the logfile) and still returns `True`, so the caller counts the trajectory as added → silent data loss. Log it and return `False`:

```suggestion
        try:
            self.dbhandle.execute(sql_str)
        except Exception as e:
            log.warning(f'failed to add trajectory to {tblname}: {e}')
            log.warning(sql_str)
            return False
        self.dbhandle.commit()
        return True
```

wmpl/Trajectory/CorrelateDB.py
```diff
@@ -0,0 +1,1092 @@
...
+            archdb.closeObsDatabase()
+
+            # attach the arch db, copy the records then delete them
+            archdb_fullname = os.path.join(db_path, f'{archdb_name}')
+            self.dbhandle.execute(f"attach database '{archdb_fullname}' as archdb")
+            try:
+                self.dbhandle.execute(f'insert or replace into archdb.paired_obs select * from paired_obs where obs_dt < {archdate_jd}')
+            except Exception:
+                log.warning('unable to archive observations database')
+                purge_ok = False
```
dvida: **Bug: the attached database is never detached.**

The `merge*` functions detach, but the three `archive*` functions do not. On a second archive call on the same connection SQLite raises `database archdb is already in use`, outside any try. Add a `finally` that detaches:

```suggestion
            self.dbhandle.execute(f"attach database '{archdb_fullname}' as archdb")
            try:
                self.dbhandle.execute(f'insert or replace into archdb.paired_obs select * from paired_obs where obs_dt < {archdate_jd}')
            except Exception:
                log.warning('unable to archive observations database')
                purge_ok = False
            finally:
                self.dbhandle.execute("detach database 'archdb'")
```

wmpl/Trajectory/CorrelateDB.py
```diff
@@ -0,0 +1,1092 @@
...
+            cur = self.dbhandle.execute(f"attach database '{archdb_fullname}' as archdb")
+            for table_name in ['trajectories', 'failed_trajectories']:
+                try:
+                    # bulk-copy if possible
+                    cur.execute(f'insert or replace into archdb.{table_name} select * from {table_name} where jdt_ref < {archdate_jd}')
+                except Exception:
+                    log.warning(f'unable to archive {table_name} in trajectories database')
+                    purge_ok = False
+
+            self.dbhandle.commit()
```
dvida: **Same missing-detach bug as in `archiveObsDatabase`** (see the observations version). Detach after committing:

```suggestion
            cur = self.dbhandle.execute(f"attach database '{archdb_fullname}' as archdb")
            for table_name in ['trajectories', 'failed_trajectories']:
                try:
                    # bulk-copy if possible
                    cur.execute(f'insert or replace into archdb.{table_name} select * from {table_name} where jdt_ref < {archdate_jd}')
                except Exception:
                    log.warning(f'unable to archive {table_name} in trajectories database')
                    purge_ok = False

            self.dbhandle.commit()
            cur.execute("detach database 'archdb'")
```

wmpl/Trajectory/CorrelateDB.py
```diff
@@ -0,0 +1,1092 @@
...
+        if archdate_jd is None:
+            keep_dt = datetime.datetime.now().replace(tzinfo=datetime.timezone.utc) - datetime.timedelta(days=21)
+        else:
+            keep_dt = jd2Date(archdate_jd,dt_obj=True)
+
+        purge_ok = True
+        if arch_prefix:
+            # create the archive database if it doesnt exist
+            archdb_name = f'{arch_prefix}_candidates.db'
+            archdb = CandidateDatabase(db_path, archdb_name)
```
dvida: **Bug: opening the archive with the default `keep=21` makes it purge its own old rows.**

`CandidateDatabase.__init__` runs `purgeCandDatabase` whenever `keep > 0`, so every time this archive DB is opened it deletes anything older than 21 days — i.e. the archive can never actually hold old data. Open it with purging disabled:

```suggestion
            archdb = CandidateDatabase(db_path, archdb_name, keep=0)
```

wmpl/Trajectory/CorrelateDB.py
```diff
@@ -0,0 +1,1092 @@
...
+            archdb.closeCandDatabase()
+
+            # attach the arch db, copy the records then delete them
+            archdb_fullname = os.path.join(db_path, f'{archdb_name}')
+            cur = self.dbhandle.execute(f"attach database '{archdb_fullname}' as archdb")
+            try:
+                cur.execute(f'insert or replace into archdb.candidates select * from candidates where ref_dt < {keep_dt.timestamp()}')
+            except Exception:
+                log.warning(f'unable to archive candidate database')
+                purge_ok = False
```
dvida: **Same missing-detach bug here** (see `archiveObsDatabase`). Also drops the stray `f` prefix on the non-f-string at line 900 (flagged by pyflakes):

```suggestion
            cur = self.dbhandle.execute(f"attach database '{archdb_fullname}' as archdb")
            try:
                cur.execute(f'insert or replace into archdb.candidates select * from candidates where ref_dt < {keep_dt.timestamp()}')
            except Exception:
                log.warning('unable to archive candidate database')
                purge_ok = False
            finally:
                cur.execute("detach database 'archdb'")
```

wmpl/Trajectory/CorrelateDB.py
```diff
@@ -0,0 +1,1092 @@
...
+                        f"'{json.dumps(traj_reduced.state_vect_mini)}',"
+                        f"{traj_reduced.phase_1_only},{traj_reduced.v_init},{traj_reduced.gravity_factor},"
+                        f"{traj_reduced.v0z},{traj_reduced.v_avg},"
+                        f"{traj_reduced.rbeg_jd},{traj_reduced.rend_jd},"
+                        f"{traj_reduced.rbeg_lat},{traj_reduced.rbeg_lon},{traj_reduced.rbeg_ele},"
+                        f"{traj_reduced.rend_lat},{traj_reduced.rend_lon},{traj_reduced.rend_ele},"
+                        f"'{json.dumps(obs_ids)}',"
+                        f"'{json.dumps(ign_obs_ids)}',1)")
+
+        sql_str = sql_str.replace('nan','"NaN"')
```
dvida: **Suggestion (robustness, not a one-line fix): use parameterised queries instead of building SQL with f-strings.**

All the SQL in this module interpolates values directly. Today the IDs are internally generated so it works, but two concrete hazards already exist:

- An `obs_id` containing a single quote raises `OperationalError` (e.g. `checkObsPaired`). I hit this in a smoke test.
- This line, `sql_str.replace('nan','"NaN"')`, corrupts **any** value whose text merely contains `nan` (a station code, a path), not just real `NaN` floats.

Switching inserts/selects to `execute("... WHERE x=?", (value,))` removes this whole class of problem. Not blocking, but worth doing while the schema is new.

wmpl/Utils/remoteDataHandling.py
```diff
@@ -23,176 +23,376 @@
...
-    """
-
-    ftpcli, remote_dir, sshcli = getSFTPConnection(remotehost) 
-    if ftpcli is None:
+        self.ssh_client = None
+        self.sftp_client = None
+        
+        cfg = ConfigParser()
+        cfg.read(cfg_file)
+        self.mode = cfg['mode']['mode'].lower()
```
dvida, line 82: **Robustness: a config file missing the `[mode]` section raises an uncaught `KeyError` instead of just disabling remote processing.**

`ConfigParser.has_option` returns False if either the section or the key is missing, so this fails gracefully the same way the invalid-mode path below already does:

```suggestion
        cfg.read(cfg_file)
        if not cfg.has_option('mode', 'mode'):
            log.warning('remote cfg: [mode] section/key missing, not enabling remote processing')
            self.mode = 'none'
            return
        self.mode = cfg['mode']['mode'].lower()
```

wmpl/Utils/remoteDataHandling.py
```diff
@@ -23,176 +23,376 @@
...
+        if not self.initialised or not self.getSFTPConnection(verbose=verbose):
+            return False
+
+        for pth in ['files', 'files/candidates', 'files/phase1', 'files/trajectories', 
+                    'files/candidates/processed','files/phase1/processed']:
+            try:
+                self.sftp_client.mkdir(pth)
+            except Exception:
+                pass
+            self.sftp_client.chmod(pth, 0o777)
```
dvida: **Robustness: only `mkdir` is inside the try, but `chmod` can also fail** (dir owned by another user, dropped connection). If it raises here the exception escapes `collectRemoteData`, skips `closeSFTPConnection`, and crashes the child. Guard it too (the sibling code in `uploadToMaster` already does):

```suggestion
        for pth in ['files', 'files/candidates', 'files/phase1', 'files/trajectories', 
                    'files/candidates/processed','files/phase1/processed']:
            try:
                self.sftp_client.mkdir(pth)
            except Exception:
                pass
            try:
                self.sftp_client.chmod(pth, 0o777)
            except Exception:
                pass
```

wmpl/Utils/remoteDataHandling.py
```diff
@@ -23,176 +23,376 @@
...
+                fullname = f'{rem_dir}/{trajfile}'
+                processed_name = f'{rem_dir}/processed/{trajfile}'
+                localname = os.path.join(local_dir, trajfile)
+                if verbose:
+                    log.info(f'downloading {fullname} to {localname}')
+
+                res = self.getWithRetry(fullname, localname)
+                if res:
+                    num_received += 1
+                    self.renameWithRetry(fullname, processed_name)
```
dvida: **The result of `renameWithRetry` is ignored.** If the move to `processed/` fails after a successful download, the file stays in `candidates/` and gets downloaded and solved again next pass (duplicate work, and it can be handed to another node). At least surface it:

```suggestion
                res = self.getWithRetry(fullname, localname)
                if res:
                    num_received += 1
                    if not self.renameWithRetry(fullname, processed_name):
                        log.warning(f'downloaded {fullname} but failed to move it to processed/ '
                                    f'- it may be downloaded and solved again next pass')
```

(Optional stronger fix: if the rename fails, delete the local copy so this node doesn't solve it — leave it for another node/pass.)

requirements.txt
```diff
@@ -19,4 +19,4 @@ watchdog==3.0.0; python_version == '3.7'
 watchdog; python_version >= '3.8'
 paramiko
 rebound==4.3.0
-reboundx==4.3.0 ; sys.platform != 'win32'
+#reboundx==4.3.0 ; sys.platform != 'win32'
```
dvida: **Was this intentional?** `reboundx` backs the REBOUND gravity/orbit path, so commenting it out drops it from every fresh `pip install -r requirements.txt`. It also looks unrelated to distributed processing — possibly a local build workaround that got committed. If it needs to stay disabled for a platform, better to keep the dependency and document the workaround:

```suggestion
reboundx==4.3.0 ; sys.platform != 'win32'
```

wmpl/Trajectory/Trajectory.py
```diff
@@ -64,6 +65,18 @@ def njit(func, *args, **kwargs):
...
+log = logging.getLogger("traj_correlator")
+
+# If no handlers are configured, add a simple console log handler
+# This will be overridden if the calling app sets up its own logger.
+if len(log.handlers) == 0:
+    console_handler = logging.StreamHandler()
+    log_formatter = logging.Formatter()
+    console_handler.setFormatter(log_formatter)
+    log.addHandler(console_handler)
+    log.setLevel(logging.DEBUG)
```
dvida: **Heads-up (affects all WMPL users, not just the correlator): a library module shouldn't attach a handler and force DEBUG at import.**

Because this runs on import of `Trajectory`, every user who imports it gets: previous `print` output now going to **stderr** instead of stdout, the `traj_correlator` logger pinned to **DEBUG** regardless of their own config, and possible double-logging if their app configures the root logger later.

The usual library pattern is `log.addHandler(logging.NullHandler())` and no `setLevel` — let the calling app decide handlers/level (CorrelateRMS already sets up its own). Flagging rather than hard-suggesting since the standalone correlator relies on seeing this output.

wmpl/Trajectory/Trajectory.py
```diff
@@ -3410,14 +3421,15 @@ def estimateTimingAndVelocity(self, observations, weights, estimate_timing_vel=T
 
             ### Try different methods of optimization until it is successful ##
 
-            # If there are more than 5 stations, use the advanced L-BFGS-B method by default
-            if len(self.observations) >= 5:
+            # If there are more than seven stations, use the advanced L-BFGS-B method by default
+            # - threshold increased from five as modern hardware can handle it fine
+            if len(self.observations) >= 7:
```
dvida: **Behaviour change that reaches every WMPL user, and isn't gated behind a flag.** This raises the timing-optimizer threshold from `>= 5` to `>= 7` stations. Events with 5 or 6 stations now take a different optimizer path (L-BFGS-B vs the `SLSQP/TNC/None` sequence), so their timing/velocity/trajectory results can change for anyone using `Trajectory.py`, not only the correlator. If that's intended, worth calling out in the changelog; if you want the correlator-only, gate it behind a parameter.

wmpl/Trajectory/CorrelateRMS.py
```diff
@@ -934,22 +888,18 @@ def loadUnpairedObservations(self, processing_list, dt_range=None):
                     continue
 
                 # Add only unpaired observations
-                if not self.db.checkObsIfPaired(met_obs):
-
+                if not self.checkIfObsPaired(met_obs.id, verbose=verbose):
```
dvida: **Bug: `verbose` is undefined here.** `loadUnpairedObservations(self, processing_list, dt_range=None)` has no `verbose` parameter, so this resolves only to the `__main__` global and raises `NameError` when the class is used as a library. `self.verbose` is what's intended (it's set in `__init__`):

```suggestion
                if not self.checkIfObsPaired(met_obs.id, verbose=self.verbose):
```

(There's at least one more `verbose=verbose` in a method without that parameter — worth grepping for the pattern.)

wmpl/Trajectory/CorrelateRMS.py
```diff
@@ -1069,108 +1018,175 @@ def trajectoryFileInDtRange(self, file_name, dt_range=None):
...
+                    log.info(f'   - removing duplicate trajectory {remove_id} from database')
+                    self.trajectory_db.removeTrajectoryById(remove_id)
+
+                    # only delete the disk file if they're in different physical locations
+                    if rw.traj_file_path != rw.traj_path_next:
+                        log.info(f'    - and removing files from {remove_path}')
+                        shutil.rmtree(os.path.split(remove_path)[0])
+
+                # remove the row from the dataframe to avoid reprocessing it
+                traj_df.drop(idx)
```
dvida: **Bug: this is a no-op.** `DataFrame.drop` returns a new frame by default, so the row is never actually removed and the comment's intent ('avoid reprocessing it') doesn't happen:

```suggestion
                traj_df.drop(idx, inplace=True)
```

wmpl/Trajectory/CorrelateRMS.py
```diff
@@ -1518,53 +1473,53 @@ def removeTrajectory(self, traj_reduced):
...
-        else:
-            log.warning(f'unable to find _processing file {pick}')
-        if not success:
-            # save the pickle in case we get new data later and can solve it
-            savePickle(traj, os.path.join(self.phase1_dir, 'processed'), fldr_name + '_trajectory.pickle')
-        return 
-
+        jdt_ref = min([obs.jdt_ref for obs, _, _ in candidate])
+        stations = [obs.station_id for obs, _, _ in candidate]
+        return self.trajectory_db.checkCandIfFailed(jdt_ref, stations)
```
dvida: **Bug: `TrajectoryDatabase` has no `checkCandIfFailed` method**, so this raises `AttributeError` if ever called. It looks like dead code right now (I couldn't find a caller). Either implement the DB method (a lookup in `failed_trajectories` by jdt_ref + stations, like `checkTrajIfFailed`) or remove this method.

wmpl/Trajectory/CorrelateRMS.py
```diff
@@ -1776,25 +2069,68 @@ def _breakHandler(signum, frame):
         help="Use best N stations in the solution (default is use 15 stations).")
 
     arg_parser.add_argument('--mcmode', '--mcmode', type=int, default=0,
```
dvida: **Minor:** the new options pass the same flag string twice, e.g. `add_argument('--mcmode', '--mcmode', ...)`, which makes `--help` print `--mcmode MCMODE, --mcmode MCMODE`. Applies to `--mcmode`, `--archivemonths`, `--maxtrajs`, `--autofreq`, `--verbose`, `--addlogsuffix` (2071–2085). Drop the duplicate string in each.

Documentation/WMPL_Upgrades_2026April.md
```diff
@@ -0,0 +1,220 @@
...
+
+**Child Node Configuration**
+
+The child must be running in mode 1 or 2 - no other mode is supported at present - and should be run with the `-a --autofreq 15` flags, to ensure that it continues to collect and process data promptly. 
+
+The child configuration file specifies the server, user and key to use for connections to the parent node. Port is optional but can be specified if a non-standard SFTP port is in use.
+
+\[mode\]  
+mode = child  
+<br/>\[children\]
```
dvida: **The documented section name is wrong** — following this verbatim silently disables remote processing. The code reads `host/user/key/port` from the `[sftp]` section (`remoteDataHandling.py` ~L84-94), not `[children]`:

```suggestion
<br/>\[sftp\]  
```

Documentation/Processing_High_Data_Volumes.md
```diff
@@ -0,0 +1,146 @@
...
+- run the solver on the candidates
+- upload any successful solutions to the parent node. 
+
+Child nodes do not need access to the raw data and do not require any long-term storage. All solved trajectories are uploaded back to the parent node and then removed from the child. Transient data on the child are housekept after a few days. 
+
+## 1: Setup Connectivity
+### 1a: Create Child Node Public Keys
+Login on each child node as the account that will run WMPL *on that node* and create an SSH key, then copy it to the parent:
+``` bash
+ssh-keygen -t ed25519 -f ~/.ssh/wmpl -N ""
```
dvida, line 38: **This key type won't work with the current code.** `remoteDataHandling.py` (~L116) hardcodes `paramiko.RSAKey.from_private_key_file`, so an ed25519 key fails to load and the child crashes at connect. Either document an RSA key:


```suggestion
ssh-keygen -t rsa -b 4096 -f ~/.ssh/wmpl -N ""
```

Better still would be to fix the code to auto-detect the key type (RSA/Ed25519/ECDSA) so ed25519 works as documented.

Documentation/Processing_High_Data_Volumes.md
```diff
@@ -0,0 +1,146 @@
...
+sudo chmod -R 775 /home/$NODENAME/files
+```
+You should end up with a structure like the one below. 
+
+![image](node_structure.png)
+
+### 1e: Set Child Node ACLs on the Parent
+Group permissions may not be sufficient to ensure the parent account can access the child accounts. To make certain you can apply extended ACLs to the folders as follows:
+```bash
+sudo setfacl -R user:ubuntu:rwx ~node1/files
```
dvida: **This `setfacl` command is missing `-m`** and errors out as written:

```suggestion
sudo setfacl -R -m u:ubuntu:rwx ~node1/files
```

dvida:
## Follow-up review — inline fixes + a few bigger design items

I went back over the current head and left **20 inline comments**, most with one-click `suggestion` blocks for the mechanical fixes (missing `status=1` filter, the `getTrajectories` crash branch, swallowed insert errors, the `attach`-without-`detach` in all three archive functions, the archive-purges-itself bug, the `verbose`/`traj_df.drop`/config-KeyError/chmod/rename issues, plus the doc and requirements items). All six of the earlier Codex findings are genuinely fixed — thanks.

Three things below need a design decision rather than a one-liner, so I've put them here instead of as suggestions.

### 1. Millisecond timestamp drift → duplicate trajectories (your known issue — root cause)
A trajectory's identity is its output folder, named from `jdt_ref` down to the millisecond (`generateTrajOutputDirectoryPath`, ~L1376), and the DB key is `jdt_ref REAL UNIQUE`. On a **single machine** this is handled: after a phase2 MC re-solve shifts `jdt_ref`, `solveTrajectory` (CorrelateEngine ~L1067-1071) notices and calls `removeTrajectory(orig_traj, remove_phase1=True)`, which deletes the stale phase1 pickle, the old folder, **and the old DB row**.

Distributed, that last step can't happen: phase2 runs on a mode-2 child whose `removeTrajectory` path (CorrelateRMS L1466-1476) only deletes local folders and `return`s **before touching the DB** (the child has no master DB). The new phase2 result has a *different* `jdt_ref` but the *same* `traj_id`, so when it's merged into the master with `insert or replace` it lands as a **new** row — the old phase1 row survives → duplicate trajectory + orphaned phase1 pickle on the master.

Suggested direction: make phase1↔phase2 identity the stable `traj_id`, and have the master reconcile on merge — when a phase2 result arrives whose `traj_id` already exists with a different `jdt_ref`, delete the old row/folder. That removes the dependence on the timestamp staying byte-identical across machines.

### 2. Files are published under their final names while the other side is reading them
Both directions write directly to the final filename and the other side polls for it, so a reader can pick up a half-written file:

- **Parent → child:** the parent writes candidate pickles with a plain `open()+dump` / `shutil.copy` into `candidates/`, while the child SFTP-lists that same folder. A mid-write download gives a truncated pickle, which the child then moves into `processed/` — the candidate is never solved, `checkAndAddCand` stops it being regenerated, and `purgeProcessedData` deletes the good copy after a while. (The `'processing'` filter in the listdir is dead — nothing writes that suffix anymore.)
- **Child → parent:** children upload results to final names while the parent `copy`+`remove`s them and merges the uploaded `.db` files. A partially-uploaded sqlite file hits `attach database` in `mergeObsDatabase`/`mergeTrajDatabase`, which is **outside any try**, so `file is not a database` propagates up through `moveUploadedData` and crashes the parent loop — and it re-crashes every pass because the bad file is never quarantined.

Standard fix for both: write to a temp name, then `rename` to the final name (rename is atomic on the same filesystem / on the SFTP server). Also worth wrapping the DB merges in a try and moving unreadable uploads aside.

### 3. "Downloaded" is treated as "done"
A candidate is moved to `processed/` the moment the download succeeds, not when a result comes back. So if a child dies with its local copy gone, or can never upload, that candidate is silently lost with no automatic requeue (the stale-file requeue never looks in `processed/`). The common case — child crashes and restarts with its disk intact — does recover. If you want it robust, either move to `processed/` only once the *result* is uploaded, or add a periodic master check for "candidate in processed/ with no matching phase1/trajectory after N days → requeue".

### Smaller notes already inline
5→7 station optimizer threshold in `Trajectory.py` changes results for 5–6 station events for *all* users (not gated); `reboundx` got commented out of `requirements.txt`; the JSON→SQLite migration copies observations and *failed* trajectories (silently capped at 14 days) but not successful ones; and the docs' child config section (`[children]` vs `[sftp]`) and ed25519 key instructions don't match the code.

Happy to help with any of these if useful.

markmac99:
Thanks Denis, we've been away for a couple of days so i will look at your findings tomorrow (monday). 

on point 3, this was deliberate: the downloaded file _should_ be recoverable from the child, but  if its not immediately marked as processed, i run the risk of the candidate getting distributed twice (with potentially different solutions), or indeed being redistributed continuously if a child node never returns any status. The latter can happen either if a  candidate is sufficiently malformed to crash the child, or if the solution fails and no failed_traj is inserted into the fails table.  There are unfortunately a few situations where this latter happens.  

Traj_ids might help with both this and point 1 though as you mention. I'll see what i can do with this.  I agree these areas need to be thought about further.

wmpl/Utils/remoteDataHandling.py
markmac99: I agree with your stronger fix. I'll see about implementing that.

requirements.txt
markmac99: no, i don't tihnk it was ! I suspect i commented this out as reboundx wasn't available on Ubuntu 26.x. I will double check and correct it.

wmpl/Trajectory/Trajectory.py
markmac99: Thanks for hte input. I originally didn't have this code at all but hten realised it meant there was no onscreen output from the other calls to Trajectory such as in ECSV.py. I added this here because i didn't want to add logging to every module that called Trajectory, but i think either that or your suggestion is probably the correct approach.

wmpl/Utils/remoteDataHandling.py
markmac99: very good point.

wmpl/Utils/remoteDataHandling.py
markmac99: good point, will fix it.

wmpl/Trajectory/CorrelateDB.py
markmac99: I know AI bots like parameterised queries, but i personally tihnk they're less robust for a bunch of reasons learned the hard way with oracle dbs. However, i'll revisit this as you make good points and whats true for oracle might not be true for sqlite.

wmpl/Trajectory/CorrelateDB.py
markmac99: The intention was actually to purge any operational data from sqlite older than 21 days since we don't need it. 

That said, there *should* be a logic check to only purge data older than the oldest date requested with the -r parameter, so eg if you ran CorrelateRMS with  -r "(20260417-080000,20260419-080000)", then it would not purge data younger than 2026-04-17. Sounds like i might not have the logic right

Question for you though: do we want to keep candidate data older htan two or three days? We never rerun the solver over older data, and if we did, it would probably be becasue we had new data and so we'd probably need to re-identify the candidates to include new groups. Also, candidate finding is very fast, and so its not a bottleneck.

wmpl/Trajectory/CorrelateDB.py
markmac99: I will need to check but i think the master database is subsequentlly closed which effects an automatic detach.  I certainly didn't get any issues in testing over multiple days However its certainly cleaner to detach it so I'll add it in.

wmpl/Trajectory/CorrelateDB.py
markmac99: indeed. Thanks.

wmpl/Trajectory/CorrelateDB.py
markmac99: I don't think this is currently called in a way that would trigger hte issue but i agree it should be fixed!

wmpl/Trajectory/Trajectory.py
markmac99: I am pretty sure i changed this because i found solutions that were failing with the threshold set to five, but which succeeded with the larger threshold and for which there was no performance penalty.  I agree its a change in behaviour though and i should have highlighted it.  We can change it back if you think thats the better option.

wmpl/Trajectory/CorrelateRMS.py
markmac99: Python and its secret globals :) Yes, this is an error on my part.

wmpl/Trajectory/CorrelateRMS.py
markmac99: oops, yes. Thanks.

wmpl/Trajectory/CorrelateRMS.py
markmac99: I think i removed this checkTrajIfFailed function as it never got called, but i'll fix this.

wmpl/Trajectory/CorrelateRMS.py
markmac99: Ah i wasn't aware of this. Will fix.

Documentation/WMPL_Upgrades_2026April.md
markmac99: good spot.

Documentation/Processing_High_Data_Volumes.md
markmac99: Oh thats interesting as i am sure i was using ed25519 keys. Will check and correct.

Documentation/Processing_High_Data_Volumes.md
markmac99: agreed.

markmac99:
ps - thanks for finding time to review this. When you've stared at it for as long as i have, you stop seeing the bugs :)

dvida:
Absolutely, it's a monumental piece of work and a second pair of eyes is always helpful :)

wmpl/Trajectory/CorrelateDB.py
markmac99: I've  resolved this and also fixed an issue when retrieving data from the failed_trajectories table.  This is not currently used anywhere, but worth correcting anyway.

wmpl/Trajectory/CorrelateDB.py
markmac99: resolved.

wmpl/Trajectory/CorrelateDB.py
markmac99: done

wmpl/Trajectory/CorrelateDB.py
markmac99: done

wmpl/Trajectory/CorrelateDB.py
markmac99: done

wmpl/Utils/remoteDataHandling.py
markmac99: done

wmpl/Utils/remoteDataHandling.py
markmac99: done

wmpl/Utils/remoteDataHandling.py
markmac99: done, i think

wmpl/Trajectory/CorrelateRMS.py
markmac99: done

wmpl/Trajectory/CorrelateRMS.py
markmac99: removed the function as its not used.

Documentation/WMPL_Upgrades_2026April.md
markmac99: resolved

Documentation/Processing_High_Data_Volumes.md
markmac99: resolved

Documentation/Processing_High_Data_Volumes.md
markmac99: Added support for RSAKey, ECDSAKey, Ed25519Key

wmpl/Trajectory/CorrelateRMS.py
markmac99: done

wmpl/Trajectory/CorrelateRMS.py
markmac99: fixed

wmpl/Trajectory/CorrelateDB.py
markmac99: As i thought, this function isn't actually used - i added it for symmetry with the Traj and Obs databases.  I've applied the fix anyway. 

I think in reality we don't need to archive the candidates. We do not generally reprocess historic data, and if we were to, it only would be for one of two reasons - significant new data, or an improved solver methodology. In both cases i think we'd want to start from scratch, creating new candidates and solving afresh, and so we'd want to ignore any existing candidate (and indeed paired obs) data.

requirements.txt
markmac99: Ah, i remember now. I had to exclude this on my build and test server as reboundx doesn't build on headless Ubuntu 24.04.4 LTS with Python 3.11, as shown below. I suspect there's another dependency i need. 

```
      building 'libreboundx' extension
      creating build/temp.linux-aarch64-cpython-311/src
      gcc -pthread -B /home/ubuntu/miniconda3/envs/wmpl/compiler_compat -Wsign-compare -DNDEBUG -fwrapv -O3 -Wall -fPIC -O3 -isystem /home/ubuntu/miniconda3/envs/wmpl/include -fPIC -O3 -isystem /home/ubuntu/miniconda3/envs/wmpl/include -fPIC -DLIBREBOUNDX -Isrc -I/home/ubuntu/venvs/wmpl/include -I/home/ubuntu/miniconda3/envs/wmpl/include/python3.11 -I/tmp/pip-build-env-yemt1c0z/overlay/lib/python3.11/site-packages/rebound -c src/central_force.c -o build/temp.linux-aarch64-cpython-311/src/central_force.o -fstrict-aliasing -O3 -std=c99 -Wno-unknown-pragmas -DREBXGITHASH=b940952556e2e94562a3ff1563ed5c49f788706c -DLIBREBOUNDX -D_GNU_SOURCE -fPIC
      src/central_force.c:60:10: fatal error: rebound.h: No such file or directory
         60 | #include "rebound.h"
            |          ^~~~~~~~~~~
      compilation terminated.
      error: command '/usr/bin/gcc' failed with exit code 1
```

markmac99: Confirmed, reboundx==4.3.0 isn't buildable with Python 3.11, for which you need 4.6.2. This forces an upgrade of rebound to 4.6.0. as well. 

I don't use rebound so for now i will change the constraint to 
```
reboundx==4.3.0 ; sys.platform != 'win32' and python_version != '3.11'
```
At some point the developers of  the WMPL rebound code can test with 4.6.2 (and ideally, also rename the WMPL package to avoid the name-clash) and hopefully we can lose the second constraint.

wmpl/Trajectory/Trajectory.py
markmac99: I have parameterised this as a keyword argument to Trajectory.__init__.  Default value 5 so unchanged for other callers such as ECSV and CAMS, but I'm passing 7 in from CorrelateEngine.initTrajectory.  Testing in progress.

wmpl/Trajectory/Trajectory.py
markmac99: Unfortunately, adding a null handler has the effect of suppressing all output from Trajectory when called from for example Formats.ECSV. 

I'll have to experiment to see what works.

markmac99:
@dvida  removing previous comment as i think i just figured out a solution. I'll pass an optional "parent logger" as a parameter to the constructor of Trajectory(). If thats None, then i'll create a console-only logger within Trajectory.py. If the parent logger is not None, then I'll use it. 

I think this will work, but need to test it from a few places.

wmpl/Trajectory/CorrelateDB.py
markmac99: Using qmark or named parameter syntax in Sqlite: I think i've done this too but it will need some fairly extensive testing since there seem to be differences in how sqlite handles parameterised queries vs text strings.

markmac99: done, tested and deployed yesterday to ukmon, ran fine last night without issues. 
ps I'll add my test cases to the github repo for any thoughts on what else i should test.

wmpl/Trajectory/Trajectory.py
markmac99: I added a `parentlogger` kwarg to Trajectory(), default value None. If the caller supplies a parent logger then it is assigned to a member variable of the Trajectory() class and to the global `log` variable defined at line 70.  If no parent is supplied a simple logger is created within the Trajectory class, writing to stdout. This allows all functions directly called from Trajectory() to use the logger. 
 Note that the function _MCTrajSolve that is spawned off into separate threads cannot use the logger as its local to Trajectory() and so its one message can only go to console.

markmac99:
# Update 29/Jul

I've addressed all the minor points plus the point about logging and the point about using named or qmark parameters rather than f-strings. The revised version was deployed to ukmon yesterday and ran well this morning.  Detailed update below:

## Bug: CorrelateDB: getTrajectories: 
this can return deleted trajectories, and the not jdt_end branch crashes.

### Resolution
Added status check and fixed logic to fetch rows properly.

## Bug: CorrelateDB: addTrajectories: 
a failed insert is hidden and the trajectory is reported as saved.

### Resolution
trap exception, log output and return False

## Bug: CorrelateDB: archiveTrajDatabase: 
the attached database is never detached.

### Resolution
catch and log any errors and detach the database outside the try (for all three databases)

## Bug: CorrelateDB: CandidateDatabase: 
opening the archive with the default keep=21 makes it purge its own old rows.

### Resolution
Add 'keep' parameter to purgeCandDatabase(), default value 21, with logic to suppress purge if keep == 0. Pass keep=0 in ArchiveCandDatabase() so that the archive isn't purged.

## Bug:remoteDataHandling:init
Robustness: a config file missing the [mode] section raises an uncaught KeyError instead of just disabling remote processing.

### Resolution
use has_option test to fail gracefully if section is missing

## Bug:remoteDataHandling:init sftp failures
Robustness: only mkdir is inside the try, but chmod can also fail (dir owned by another user, dropped connection). If it raises here the exception escapes collectRemoteData, skips closeSFTPConnection, and crashes the child. 

### Resolution
guarded it  with a try/except

## Bug:remoteDataHandling:renameWithRetry
The result of renameWithRetry is ignored. 

### Resolution
logged the warning and returned False if it failed after 10 retries

# Enh:remoteDataHandling:sftp key types
Only supported RSA keys, 

### Resolution
addded support for other key types

## Bug:CorrelateRMS:loadUnpairedObservations
bug: verbose is undefined here. 

###Resolution
use the member variable rather than the undefined one!

## Bug:CorrelateRMS:updateTrajectoryDatabase
DataFrame.drop returns a new frame by default, so the row is never actually removed and the comment's intent ('avoid reprocessing it') doesn't happen:

### Resolution
add 'inplace=True' to the drop statement.

## Bug:CorrelateRMS:
Bug: TrajectoryDatabase has no checkCandIfFailed method,
### Resolution
remove the unused method which was calling the nonexistent method

## Bug CorrelateRMS main
Minor: the new options pass the same flag string twice,

### Resolution
Drop the duplicate string in each.

## Bugs:Documentation Mistakes
All corrected i think. 

## Definiing a Logger
Heads-up (affects all WMPL users, not just the correlator): a library module shouldn't attach a handler and force DEBUG at import.

### Resolution
I added a `parentlogger` kwarg to Trajectory(), default value None. If the caller supplies a parent logger then it is assigned to a member variable of the Trajectory() class and to the global `log` variable defined at line 70.  If no parent is supplied a simple logger is created within the Trajectory class, writing to stdout. This allows all functions directly called from Trajectory() to use the logger. 
Note that the function _MCTrajSolve that is spawned off into separate threads cannot use the logger as its local to Trajectory() and so its one message can only go to console. 
 
I tested this from both ECSV and CorrelateRMS and it works as expected. 

## Use of f-Strings for DB Queries
All the SQL in this module interpolates values directly. Today the IDs are internally generated so it works, but two concrete hazards already exist:

### Resolution
Switch to qmark and named parameters.  See test cases added in Tests folder

## Behaviour change in timing optimisation
Behaviour change that reaches every WMPL user, and isn't gated behind a flag. 

### Resolution
gatekeep it behind a kword arg to Trajectory(), default value 5, set to 7 by the correlator


To Do

1) Millisecond time stamp movements. 
As noted, when Phase2 solutions are uploaded there is no process to check and remove any previous phase-1 solution.
This is only done if the solution was created locally. 

2) Chance of partially uploaded files being processed
As files re uploaded with their final, real name theres a chance of partial files being read/processed/moved
Solution - use temp names then rename after 

3) downloaded treated as done
Downloaded files are immediately treated as done and so if a child collects them and hten crashes, theyre not re-distributed by the parent.
The problem here is that if we do *not* mark them as processed, then they might be redistributed and processed twice.
Simlarly if we temporarily rename them to eg _processing, then since the child only uploads a list of successful solutions, the fails would be left indefinitely. Worse, if the child crashed, the temp files might be left indefinitely too. 

Possible solution: temporarily rename to _processing. 
When uploading solutions, child should move and rename its candidates, both failed and succeeded
Parent should check after say 24 hours and rename any candidates that are still marked _processing so they are picked up again. 
There's a risk that a candidate will be so bad it crashes the child, so we might also need a counter to indicate how many times a cand has been tried. After say 5 tries, we can then abandon it as a failed cand.

markmac99:
> ### 1. Millisecond timestamp drift → duplicate trajectories (your known issue — root cause)
> A trajectory's identity is its output folder, named from `jdt_ref` down to the millisecond (`generateTrajOutputDirectoryPath`, ~L1376), and the DB key is `jdt_ref REAL UNIQUE`. On a **single machine** this is handled: after a phase2 MC re-solve shifts `jdt_ref`, `solveTrajectory` (CorrelateEngine ~L1067-1071) notices and calls `removeTrajectory(orig_traj, remove_phase1=True)`, which deletes the stale phase1 pickle, the old folder, **and the old DB row**.
> 
> Distributed, that last step can't happen: phase2 runs on a mode-2 child whose `removeTrajectory` path (CorrelateRMS L1466-1476) only deletes local folders and `return`s **before touching the DB**

In fact this should be handled already elsewhere.  

Each trajectory contains a `pre_mc_longname` value, which identifies the phase1 solution. In  `moveUploadedData` we read this field and if the folder exists, we delete it:
``` python
  traj = loadPickle(src_path, src_name)
  if hasattr(traj, 'pre_mc_longname') and traj.pre_mc_longname != traj.longname:
      # When saving MC-phase solns we need to check if the MC phase has changed the reference time
      # slightly, and if so, remove the original folder to ensure no duplicates.
      pretraj = traj.pre_mc_longname
      pre_path = os.path.join(self.output_dir, 'trajectories', pretraj[:4], pretraj[:6], pretraj[:8], pretraj)
      log.info(f'removing {pre_path}')
      if os.path.isdir(pre_path):
          shutil.rmtree(pre_path, ignore_errors=True)
```
Then, in `updateTrajectoryDatabase`, at around line 1100,  we scan for trajectories with exactly the same observations, and remove whichever one isn't on disk from the database. We do remove by Traj ID, so there's a chance we'll remove both the ph1 and ph2 solns from the DB if the traj_id didn't change, but a few lines later at line 1173, we will re-add any trajs that are still on disk, so this will re-add the ph2 solns if they got inadvertently removed.  

So, while its a bit convoluted, i think it does do what we need. and in testing, I did not encounter any situations where phase1 solns were orphaned.  I will double check though.

markmac99:
> ### 2. Files are published under their final names while the other side is reading them
> Both directions write directly to the final filename and the other side polls for it, so a reader can pick up a half-written file:
> 
> * **Parent → child:** the parent writes candidate pickles with a plain `open()+dump` / `shutil.copy` into `candidates/`, while the child SFTP-lists that same folder. 
> * **Child → parent:** children upload results to final names while the parent `copy`+`remove`s them and merges the uploaded `.db` files. 

Note that this risk already exists because even on a single server the phase1 process could be writing to the `phase1/` folder while the mc-phase process is reading from it. 

Issues can arise in three places .
1) parent is distributing candidate or phase1 pickles to child node folders, while child is simultaneously downloading. 
2) parent is scanning for uploads and processing files, while child is simultaneously uploading
3) parent  phase1 process is saving phase1 solutions while the parent mc-phase is reading them. 

To ensure partly-created pickles are not read by sftp or other processes on the parent, I've altered `savePickle` to use a unique temporary name and then rename the file afterwards.  

To ensure partly-complete uploads are not read, i've altered `putWithRetry` to use a unique temporary name and then rename afterwards. 

I've also created a function `safeCopyOrMove()` in OSTools, which uses a temporary name when copying or movnig files around, and i've used it wherever there was a chance of a read/write conflict. Note that we can't directly use `os.replace()` as this doesn't work across filesystems.

markmac99:
I've made what i hope are the final changes. Candidates and phase1 solutions that are being processed are now no longer moved to the processed/ folders till after processing is complete. To prevent the same data being picked up by another process or remote node, the status is marked as 2 in the database. This entailed some changes to the DB class to handle concurrency better. 

Testing:
I ran three instances of the solver on my server, one in each mode, plus two remote nodes one in Phase1 mode and one in MC mode. Using 25 days of data from 13 cameras, i ran the solvers continuously, introducing a new day's data at random intervals between 10 and 30 minutes apart.  The test ran over an 18-hour window without issue.

The parent node distributed and collected data as expected as well as processing is own workload. The remote nodes collected data, processed and uploaded as expected. There was occasionally an upload failure sending data to the parent node but this is handled by sending the data on the next pass.  

Let me know if you spot anything else i need to look at.

dvida:
@markmac99 — I brought this branch up to date with master and pushed the result (`9df8139..10a3fb1`). Summary of what changed and why, so nothing lands as a surprise.

## 1. Merge with master (`d8a511d`)

Master was 71 commits ahead and three files conflicted. All three resolutions are mechanical:

- **`requirements.txt`** takes master's version. The `rebound==4.3.0` / `reboundx==4.3.0 ; python_version != '3.11'` pin turned out to be uninstallable in a clean venv: reboundx has no wheels, its `pyproject.toml` build-requires `rebound >= 4.0.0`, and pip's build isolation resolves that to the newest rebound whatever is pinned, so the build fails on a missing `rebound.h`. Master unpinned both (with a comment explaining why), which makes your 3.11 marker unnecessary.
- **`CorrelateEngine.py` / `CorrelateRMS.py`** keep your code. Master only touched them in the repo-wide codespell sweep (14 comment/docstring/log-string lines), so I re-applied those spellings on top of your version and checked the Python token streams: no executable token differs from your branch.
- `Trajectory.py`, `Math.py`, `OSTools.py` merged cleanly; master's changes to them in this range are comments only.

## 2. Defects fixed

None of these were covered by a test (reverting any of them left the suite green), and none was in the earlier review threads.

| Where | Problem | Fix |
| --- | --- | --- |
| `Pickling.savePickle` | Temp name was `file_name.replace('.p', '_x')`. For a name with no `.p` the temp name *is* the target, so the just-written file was removed and the rename raised — `savePickle(obj, d, "results.dat")` lost the data. Also non-atomic. | Extensionless uuid temp name (still never matches your `*.pickle*` globs) + `os.replace`, which is atomic on POSIX and Windows. |
| `Trajectory.__init__` | `global log; log = parentlogger` rebinds the module logger, so the last Trajectory built decides where *every* Trajectory in the process logs. Without a parent logger, a new `StreamHandler` was added on every construction (6 trajectories → 6 handlers → each line printed 6 times). | Store only `logger_name`; `self.log` is a read-only property resolving it. Console handler attached once. The `parentlogger=` interface is unchanged. |
| `loadPickle` | Older pickles have no `l_bfgs_b_cutoff`, which `estimateTimingAndVelocity` reads. | Back-filled with the default 5, next to `gravity_factor` / `v0z`. |
| `purgeTrajDatabase` | `return True` sat inside the `for table_name in [...]` loop, so `failed_trajectories` was never purged — and since archive = copy then purge, failed rows were re-copied into the archive every run. | Return after the loop. |
| `TrajectoryDatabase.safeDetachDatabase` | Ignored its argument, always detached `'archdb'`, so `sourcedb` stayed attached after a failed merge and the next merge could not attach it. | Detach `{dbname}`, as `ObservationsDatabase` already does. |
| `CandidateDatabase.safeDetachDatabase` | Detached the literal string `'dbname'`. | Same. |
| `clearProcessingFlag` (both classes) | No return on success, so the UPDATE ran 10× and `False` was always returned; the generic `except` logged an unbound `e` (`NameError` inside the handler). | `return True` after commit; bind the exception. |

**One of my own first attempts was wrong, and your end-to-end script caught it.** I initially stored the logger object on the instance. `toJson()` deep-copies and JSON-encodes the whole `__dict__`, so every call raised `cannot pickle '_thread.RLock' object`, and `run_correlator.sh` produced a reproducibly different trajectory set. Storing only the name (the property above) fixed it. Worth knowing if anyone later adds non-serialisable attributes to `Trajectory`.

## 3. Tests

**Three of the existing tests were already failing on the branch before I changed anything:** `test_CandDb`, `test_archAndPurgeCandDb`, `test_mergeCandDb`. `CandidateDatabase(keep=21)` purges rows older than 21 days *on open*, and the fixtures are dated 20 July, so the expected counts stopped holding once the calendar moved past them. The tests now open with `keep=0`, and a new `test_CandDbKeepPurgesOnOpen` exercises the purge deliberately with a window that holds whenever it is run.

New tests, each checked to fail when its fix is reverted:

- `test_CorrelateDB.py`: purge clears both tables; `clearProcessingFlag` resets status 2→1 and returns True; a merge from a mismatched source detaches and the next merge still works; `safeDetachDatabase` detaches both `sourcedb` and `archdb` in both classes. **17 → 23 tests, all passing.**
- `test_PickleCompat.py`: `savePickle` round-trips `results.dat`, `a.pre.pickle`, `noextension` and leaves nothing behind; the `l_bfgs_b_cutoff` back-fill.
- New `test_TrajectoryLogging.py`: parent logger doesn't leak onto the module or other instances; handlers don't accumulate; the logger is not in `__dict__` and a trajectory survives `toJson`, `deepcopy` and `pickle`; a legacy trajectory without `logger_name` still logs.

Full suite: **288 passed** (master alone: 258).

**`run_correlator.sh`**: after the fix it matched `expected_results.txt` exactly in two consecutive runs; a third showed only the documented borderline `20260712_225600.836_UK` flake (MC moved `jdt_ref` from `.836` to `.837`). Monte Carlo isn't seeded on this path, so that variation is inherent to the script, as its own note says.

## 4. Documentation

The correlator had no page in `wmpl/Docs/` at all, and this branch added its docs in a new top-level `Documentation/` folder that nothing linked to. I rewrote the content as **`wmpl/Docs/Correlator.md`** in the same style as the other manuals, covering the whole tool rather than only the new parts: the `--mcmode` bitmask (and that combined modes keep intermediates in memory), what it reads and writes, the full CLI, single-server continuous running, child-node setup, `wmpl_remote.cfg`, the three databases and their `status` values, the JSON migration and `CorrelateDB --action copy`, worked examples and troubleshooting. `node_structure.png` became an ASCII tree. It's linked from `wmpl/Docs/README.md`, the root README and `Trajectory.md`, and `Documentation/` is removed.

The `--remotehost` removal is called out as a breaking change, and `--archivemonths 0` as purge-not-archive. Changelog framing, first-person asides and the open note about merging mergeable trajectories were dropped — please check I haven't lost anything you consider operationally important.

`.gitignore` gained its missing trailing newline and a `wmpl_remote.cfg` entry.

## Left as they are, flagging for Denis

These are your design decisions and I didn't change them, but they reach beyond the correlator:

- `CorrelateEngine.initTrajectory` passes `l_bfgs_b_cutoff=7`, moving the timing-fit switch from 5 to 7 stations for every correlator solve.
- The `print` → `logging` conversion in `Trajectory.py` changes output for every library user (now without the handler duplication).
- `generateDatetimeBins` no longer snaps sub-day bins to the UTC break hour.

The branch merges cleanly with current master. Given 300+ commits including several upstream merges, a **squash merge** would keep master's history readable.

dvida:
Follow-up `b6f4f09`, docstring only: `generateDatetimeBins` now says the `utc_hour_break` snapping applies only to bins of a day or longer. I checked the change itself: master's version can't produce sub-day bins at all (6-hour bins collapse onto 12:00 UTC as zero-length or backwards bins), 1-day and 7-day output is identical to master, and the only callers are the correlator's. So correcting my earlier note, it doesn't reach any other code.

markmac99:
Ah, we'll have to disagree on a few of the spelling changes as some of them are US vs UK english :) I will forever spell useable with an e :)  Generally i avoid changing spelling as it adds noise to commits (unless one is making a commit of *only* spelling corrections). 

The documentation is great, much welcomed. and looks good. 

`Correlator.md` line 115 implies that a trajectory folder will have a station count appended. i think thats incorrect? `20260712_221506.294_UK_4sta` should be `20260712_221506.294_UK`

In `Trajectory.md` i would suggest mentioning the `-w` parameter as its extrmely useful to walk subdirectories. Several of the solvers use it. 

The corrections in CorrelateDB look good, obviously my test cases were insufficient!

markmac99:
As an aside: the docstring in generateDatetimeBins reminded me: 
"The bin edges will be at utc_hour_break UTC. 12:00 UTC is chosen because at that time it is midnight at the International Date Line, and it is very unlikely that there are any meteor cameras there, so a night of observations is never split between two bins."

This assumption is incorrect! In testing, i found several detections in AU where some cameras folders were timestamped before 12:00 UTC and others timestamped just after 12:00 UTC, and which ended up in different buckets because of the above assumption.  This arises because Aus is 8-10 hours ahead of UT so at certain times of year, detection might start very close to 12:00 UT. 
I worked around this by a change at `CorrelateRMS.py:825` in master (line 758 in the PR version), which temporarily includes the prior day's folders in the scan for new data. I'm not sure this is a full fix however.

markmac99:
I ran a test of distributed mode last night using my PC and a Linux server - no issues. 

The Linux server was dunning candidate, simple and mc mode, distributing both candidates and simple solns. 
The PC was running two separate instances one for simple and one for mc.  
I pulled in data from fourteen UKmon cameras, covering 17-22 Sept, pulling one day at a time then waiting ~10 mins before pulling the next, to simulate data arriving at various times. 

For comparison i ran the solver in single-server mode (mcmode=5) on the same dataset and got the same output. 

The data contained
- 1617 observations of which 
- 722 paired observations were grouped into 
- 360 candidate groups leading to 
- 299 successful solutions

- 94 candidates were distributed to the PC, of which 78 solved.
- 116 simple solutions were distributed to the PC of which all were solved (as expected). 

However, i did hit an interesting case where due to timing, the same phase1 cand was distributed to both parent and child nodes. This sometimes resulted in two solutions with different on-disk locations but the same traj_id.  I think i have a fix for this, currently testing.

PR #63, opened 2026-06-08
Enhance alphaBetaMasses with velocity-dependent final mass computation

AstroEloy:
This update extends alphaBetaMasses() by supporting the full alpha-beta final-mass solution when both initial and terminal velocities are provided. The previous asymptotic approximation is retained for backward compatibility when velocity information is unavailable.

The new functionality has also been integrated into the command-line interface, using the terminal velocity predicted by the fitted alpha-beta solution to ensure consistency when calculating masses with the general alpha-beta formulation.

Additional improvements include:

* Input validation for alpha, beta, slope, mu, density, drag parameter, shape coefficient, and velocity inputs.
* Optional verbose output summarizing the adopted parameters and resulting mass estimates.
* Expanded documentation and updated docstrings.
* alphaBetaVelocity() now accepts both scalar and array height inputs.
* Minor typos and documentation issues corrected.
* The meteorite-dropping possibility plot now displays the estimated final masses for both μ = 0 and μ = 2/3.
* Initial and final masses are now printed using scientific notation (:.2e) instead of fixed-point formatting.

### DIFF
--- a/wmpl/Utils/AlphaBeta.py
+++ b/wmpl/Utils/AlphaBeta.py
@@ -285,7 +285,6 @@ def lagFitVelocity(time_data, lag_data, vel_data, v0):
     return vel_fit, fit_params
 
 
-
 def minimizeAlphaBeta(v_normed, ht_normed):
     """ initiates and calls the Q4 minimisation given in Gritsevich 2007 -
         'Validity of the photometric formula for estimating the mass of a fireball projectile'
@@ -334,7 +333,6 @@ def minimizeAlphaBeta(v_normed, ht_normed):
     return res.x
 
 
-
 def fitAlphaBeta(v_data, ht_data, v_init=None):
     """ Fit the alpha and beta parameters to the given velocity and height data. 
     
@@ -349,7 +347,7 @@ def fitAlphaBeta(v_data, ht_data, v_init=None):
     Return:
         (v_init, alpha, beta):
             - v_init: [float] Input or derived initial velocity (m/s).
-            - alpha: [float] Balistic coefficient.
+            - alpha: [float] Ballistic coefficient.
             - beta: [float] Mass loss.
     """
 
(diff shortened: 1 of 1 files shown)


dvida:
Thanks for the PR — the velocity-dependent final-mass extension is a useful addition and the underlying math is correct (it's the standard Gritsevich dimensionless mass equation, with the old `exp(-beta/(1-mu))` recovered as the `v_end → 0` limit). The input validation is also a nice touch. A few things to fix before merge:

**1. Duplicate function definition (blocking).** The new `alphaBetaMasses` is *appended* after the existing one rather than replacing it. The file now has two functions with the same name (old at lines ~456–487, new right after). Python silently uses the last definition, so the old one becomes dead code. Please delete the original `def alphaBetaMasses(...)` block and keep only the new version. As-is, this also leaves two conflicting docstrings (`gamma` described as "Drag coefficient (2*drag factor)" in the old one vs. "Γ = C_D/2" in the new one).

**2. Feature isn't wired into the CLI.** The only callers are in the `__main__` block (lines ~612 and ~614), and they don't pass `vel_init`/`vel_end` — so the new full solution is never exercised by the script itself, only via direct import. Since `v_init` is already available (line ~608) and the trajectory has the final velocity, please pass them through so the CLI actually uses the new path (or note explicitly that it's library-only).

**3. Minor style.** Trailing whitespace on several blank lines (docstring opener, before `if verbose`, etc.), and add a second blank line before the `def` per PEP8.

Once the duplicate is removed and the velocities are plumbed into the main block, this is good to go.

AstroEloy:
Thanks for the review.

The duplicate alphaBetaMasses definition has been removed and only the updated implementation remains.

The CLI has been updated to use the terminal velocity predicted by the fitted alpha-beta solution and now passes vel_init and vel_end to alphaBetaMasses, so the full formulation is exercised directly from the command-line tool.

Additional updates include:
- Input validation for alpha, beta, slope, mu, density, drag parameter, shape coefficient, and velocity inputs.
- Optional verbose output summarizing the adopted parameters and resulting mass estimates.
- Expanded documentation and updated docstrings.
- alphaBetaVelocity() now accepts both scalar and array height inputs.
- Minor typos and documentation issues corrected.
- The meteorite-dropping possibility plot now displays the estimated final masses for both μ = 0 and μ = 2/3.
- Initial and final masses are now printed using scientific notation.

PR #64, opened 2026-06-12
add ReVelle & Ceplecha (2001) luminous efficiency and variable-velocity photometric mass

AstroEloy:
Add a velocity-dependent luminous efficiency option to calcMass() and make the
array-velocity case integrate the mass-loss rate properly, while keeping the
existing scalar/float behavior backward compatible.

Changes
- Add luminous_efficiency_RC2001(v) following Borovicka et al. (2022):
  panchromatic differential luminous efficiency of ReVelle & Ceplecha (2001),
  velocity terms of Eq. (8) for PE type-I bolides (c1 = +0.466, c2 = -1.538).
  Input velocity in m/s (converted to km/s internally); returns tau as a ratio.
  tau is paired with the matching  zero-magnitude power (e.g. V-band P_0m = 1500 W).

- calcMass(): tau now accepts either a float (unchanged, default 0.007) or the
  string 'rc2001', in which case tau is computed from velocity on the fly via
  luminous_efficiency_RC2001(). Unknown model names raise ValueError.

- calcMass(): for an array velocity, integrate the instantaneous mass-loss
  rate M = integral(2*I/(tau*v^2) dt), keeping tau and v inside the
  integral. The scalar (average-velocity) path is unchanged; both agree
  exactly when v and tau are constant.

- Extract calcIntensity(mag_abs, P_0m): the radiated-power expression
  I = P_0m*10^(-0.4*M) now shared by calcRadiatedEnergy() and calcMass().
  No numerical change to calcRadiatedEnergy().

Behavior change
- For an array velocity, calcMass() now returns a single scalar mass (the
  integrated value) instead of a per-point array. This matches the previous
  documented [float] return and the physics; callers that relied on the previous
  per-point array output must be updated.

### DIFF
--- a/wmpl/MetSim/MetSimErosionCyTools.pyx
+++ b/wmpl/MetSim/MetSimErosionCyTools.pyx
@@ -389,6 +389,7 @@ cpdef double luminousEfficiency(int lum_eff_type, double lum_eff, double vel, do
             5 - CAMO faint meteors
             6 - Celpecha & McCrosky (1976)
             7 - Borovicka et al. (2020) - Two strengths
+            8 - Pecina & Ceplecha (1983)
         lum_eff: [double] Value of the constant luminous efficiency (percent).
         vel: [double] Velocity (m/s).
         mass: [double] Mass (kg).
--- a/wmpl/Utils/Physics.py
+++ b/wmpl/Utils/Physics.py
@@ -73,6 +73,7 @@ def dynamicMass(bulk_density, lat, lon, height, jd, velocity, decel, gamma=1.0, 
     Return:
         dyn_mass: [float] Dynamic mass in kg.
 
+
     """
 
     # Calculate the atmosphere density at the given point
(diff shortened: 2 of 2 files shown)


dvida:
Pushed a follow-up commit (c734aa1) that restructures this to reuse the existing luminous efficiency models instead of adding a new standalone function.

**Why:** `wmpl/MetSim/MetSimErosionCyTools.luminousEfficiency()` is already the single source of truth for the lum. eff. models (codes 0–8, including RC2001 Type I/II/III). The added `luminous_efficiency_RC2001()` was mathematically identical to `lum_eff_type=1` evaluated at mass = 1 kg (the `1.15*tanh(0.38*log(mass))` term vanishes there), so it duplicated existing code and exposed only one of the available models.

**Changes:**
- Removed the standalone `luminous_efficiency_RC2001()` from `Physics.py`.
- `calcMass()` `tau` now accepts:
  - `float` → constant luminous efficiency ratio (unchanged, default 0.007 — fully backward compatible),
  - `int` → `lum_eff_type` code passed straight to the pyx `luminousEfficiency()`,
  - `str` → model name mapped via a new `LUM_EFF_MODELS` dict (e.g. `'rc2001'`, `'borovicka2020'`, …).
  This exposes all 9 models, not just RC2001.
- Velocity-dependent `tau` is computed per velocity point and the instantaneous mass-loss rate `2*I/(tau*v^2)` is integrated; scalar and array velocity share one code path.
- New `lum_eff_mass` kwarg supplies the mass for the mass-dependent models. `-1` (default) iterates mass↔tau to self-consistency.
- The pyx import is lazy, so `Physics.py` keeps no hard MetSim/Cython dependency at import time.
- Documented the previously-undocumented option 8 (Pecina & Ceplecha 1983) in the `luminousEfficiency` docstring.

**Behavior change (carried over from this PR):** for an array velocity, `calcMass()` returns a single scalar mass (the integrated value), matching the documented `[float]` return. All 8 current call sites pass scalar velocity, so they are unaffected.

**Verification:** `tau=0.007` float path is bit-for-bit unchanged; `tau='rc2001'` == `tau=1` and reproduces the previous RC2001 function at mass = 1 kg; constant array velocity agrees with the scalar path; `lum_eff_mass=-1` iteration is self-consistent with passing the converged mass explicitly; unknown model name raises `ValueError`.

Note: the pyx change is source-only — collaborators need to rebuild the Cython module for the new docstring line to appear in `__doc__` (cosmetic only).

PR #65, opened 2026-06-16
Add RC2001 deceleration correction and calcMass support

AstroEloy:
This PR improves the implementation of the luminous efficiency model from ReVelle & Ceplecha (2001) and its application in photometric mass estimation.

Changes:
- Added support for the velocity-dependent deceleration correction term described by ReVelle & Ceplecha (2001).
- Extended luminousEfficiency() with an optional v_init argument representing the pre-atmospheric velocity.
- Implemented the ReVelle & Ceplecha (2001) deceleration correction using the velocity difference (v_init - vel).
- For v_init = vel, the deceleration correction is set to zero.
- Velocity values larger than v_init are treated as physically inconsistent and will cause the logarithm evaluation to fail.
- Added documentation describing the assumptions and validity limits of the deceleration correction term.

Photometric mass calculation (calcMass):
- Added support for passing a pre-atmospheric velocity through the new v_init argument.
- Added three operating modes for the ReVelle & Ceplecha deceleration correction:
    - v_init=None: do not apply the correction (it translates into -1 for luminousEfficiency()).
    - v_init=-1: use the maximum value of the supplied velocity profile as the pre-atmospheric velocity.
    - v_init>0: use the user-provided pre-atmospheric velocity.
- Updated the documentation to clarify the expected use of velocity profiles and the assumptions behind the automatic pre-atmospheric velocity selection.

Velocity-dependent mass integration:
- Fixed the treatment of variable velocity when a constant luminous efficiency is supplied.
- The mass is now computed using
    M = \frac{2}{\tau}\int \frac{I}{v^2},dt
    when a velocity profile is provided, instead of incorrectly assuming constant velocity.

Input validation:
- Added validation that time and mag_abs have identical lengths.
- Added validation that velocity arrays have the same length as the photometric measurements.
- Added validation for lum_eff_mass, which must be either:
    - -1 (self-consistent iteration), or
    - a non-negative mass value.

Self-consistent luminous efficiency iteration:
- Added a warning message when the luminous-efficiency self-consistency iteration fails to converge after the maximum number of iterations.
- Improved documentation of the convergence procedure and mass iteration workflow.

Documentation:
- Expanded the documentation of calcMass() and luminousEfficiency().
- Clarified the distinction between constant and velocity-dependent luminous efficiency calculations.
- Added documentation for the new pre-atmospheric velocity handling and associated assumptions.

### DIFF
--- a/wmpl/MetSim/MetSimErosionCyTools.pyx
+++ b/wmpl/MetSim/MetSimErosionCyTools.pyx
@@ -376,7 +376,7 @@ cpdef double decelerationRK4(double dt, double K, double m, double rho_atm, doub
 
 
 @cython.cdivision(True) 
-cpdef double luminousEfficiency(int lum_eff_type, double lum_eff, double vel, double mass):
+cpdef double luminousEfficiency(int lum_eff_type, double lum_eff, double vel, double mass, double v_init=-1.0):
     """ Compute the luminous efficiency of the given type, velocity, and mass.
     
     Arguments:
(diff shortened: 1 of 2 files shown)


dvida:
Reviewed and verified against the original RC2001 paper (eq. 8) — the correction term and Type I/II/III constants are implemented correctly, and applying it in both velocity regimes matches the paper. Thanks Eloy!

I pushed two follow-up commits before merging:

1. **Bounded the deceleration correction** — it diverges to −∞ as (v_init − vel) → 0 (dv = 1 m/s inflates dm/dt ×24), and `log(v_init − vel)` silently returned NaN for vel > v_init rather than failing. The correction is now linearly tapered to zero below dv = 0.1 km/s and disabled for dv ≤ 0, so tau stays bounded and continuous for all inputs.
2. **Changed the automatic v_init estimate** — using the maximum of the velocity array guaranteed dv = 0 at the fastest point, putting the surrounding points into the divergent region. It's now the median velocity over the first 25% of points, which is also robust to measurement noise. Invalid v_init values (0, negatives other than -1) now raise ValueError instead of silently disabling the correction.

Verified with a synthetic-event script: exact formula match in both regimes, taper continuity, no NaN propagation, legacy call sites unchanged, all error paths raise.

PR #66, opened 2026-06-17
Add calcPE() for Ceplecha & McCrosky PE classification

AstroEloy:
Introduced a new calcPE() function implementing the Ceplecha & McCrosky (1976) PE end-height criterion for meteoroid structural classification.

Changes:
-Added computation of the PE parameter from end-height atmospheric density, initial mass, entry velocity, and radiant zenith angle.
- Added automatic support for WMPL Trajectory objects, deriving:
    - end-point atmospheric density from getAtmDensity_vect(),
    - zenith angle from the apparent radiant elevation,
    - initial velocity from traj.v_init.
- Added input validation for physically valid ranges (mass > 0, rho_e > 0, v_0 > 0, and cos(zangle) > 0).
- Implemented classification into structural groups I, II, IIIa, and IIIb following Ceplecha & McCrosky (1976).
- Added documentation emphasizing that the input mass should be the initial photometric mass computed using the Ceplecha & McCrosky (1976) luminous efficiency law to preserve consistency with the original calibration.

### DIFF
--- a/wmpl/Utils/Physics.py
+++ b/wmpl/Utils/Physics.py
@@ -255,7 +255,8 @@ def calcMass(time, mag_abs, velocity, tau=0.007, P_0m=840.0, lum_eff_mass=-1, v_
     if (lum_eff_type in [1, 2, 3, 6, 7, 8]) and (P_0m != 1500.0):
         warnings.warn("The chosen luminous efficiency model (lum_eff_type={:d}) is calibrated for "
             "a panchromatic zero-magnitude power of P_0m = 1500 W, but P_0m = {:g} W was given. "
-            "The computed mass will be scaled by the ratio of the two.".format(lum_eff_type, P_0m))
+            "No correction is applied - the returned mass will be off by a factor of {:.2f} "
+            "compared to using the calibrated value.".format(lum_eff_type, P_0m, P_0m/1500.0))
 
     # Compute the luminous efficiency using an analytical velocity-dependent function.
     import pyximport
--- a/wmpl/Utils/Physics.py
+++ b/wmpl/Utils/Physics.py
@@ -593,26 +593,23 @@ if __name__ == "__main__":
     print()
     print("Ceplecha & McCrosky (1976) PE criterion calculation:")
 
-    # Hand-checkable case: PE = -8 - 0.42*3 + 1.49*log10(15) - 1.29*log10(cos(45 deg)) = -7.31
-    pe, pe_group = calcPE(1.0, rho_e=1e-5, zangle=np.radians(45), v_0=15e3)
-    print("PE test (should return ~-7.31, IIIb): PE = {:.2f}, group = {:s}".format(pe, pe_group))
-
-    # With a 1 g mass, vertical entry and v_0 = 10 km/s, the mass and zenith angle terms vanish and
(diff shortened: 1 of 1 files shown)


dvida (commented):
Thanks Eloy! I did a thorough review of the science and the code, and pushed a polish commit (3c7a45c) directly to this branch with the fixes described below.

## Science verification

I checked the method and all numbers against Ceplecha (1988, BAICz 39, 221, Eq. 2 — which reproduces the C&M 1976 criterion verbatim), Ceplecha et al. (1998, SSRv 84), and Borovička et al. (2022, A&A 667, A158, Eq. 1). Everything checks out:

- Coefficients −0.42 / +1.49 / −1.29 — correct.
- Units (ρ_E in g/cm³, mass in g, velocity in km/s) and the SI→cgs conversions — correct.
- Group boundaries −4.60 / −5.25 / −5.70 with boundary values assigned to the weaker group — exactly as published (Ceplecha 1988, Sec. 2e).
- The note that the mass must come from the C&M (1976) luminous efficiency is right, and the `lum_eff_type=6` implementation matches Ceplecha (1988) Eq. 3 piecewise exactly.

## Fixes applied in 3c7a45c

1. **P_0m consistency (the one substantive issue).** The cm1976 luminous efficiency in `MetSimErosionCyTools.pyx` is defined for a panchromatic zero-magnitude power of 1500 W, but `calcMass()` defaults to `P_0m=840`. A mass computed with the default shifts PE by about +0.11 (−0.42·log10(840/1500)), roughly 1/6 of a group width. The docstring now says to use `calcMass(..., tau='cm1976', P_0m=1500)`.
2. **No silent override.** Passing `traj` together with explicit `rho_e`/`zangle`/`v_0` now raises a `ValueError` instead of silently ignoring the explicit values.
3. **Guarded orbit access.** A trajectory without a computed orbit now produces a clear error instead of a `TypeError` on `None`.
4. **Docstring style.** Split into `Arguments:` / `Keyword arguments:` sections to match the rest of `Physics.py`.
5. **References.** Added Ceplecha (1988) alongside C&M (1976) — the 1988 paper is the open-access source where the equation and boundaries can actually be checked.
6. **Atmosphere model note.** The original calibration used a 1970s CIRA-type atmosphere; the difference to NRLMSISE-00 at end heights is <~0.03 dex vs the 0.65 dex group widths, so this is fine — now noted in a comment.
7. **Test cases.** Added a PE block to `__main__` mirroring the `pf_table` pattern: a hand-checkable case plus one case per structural group (with a 1 g mass, vertical entry and 10 km/s, PE reduces to log10(ρ_E in g/cm³) + 1.49). All pass.

Two design choices I deliberately kept as-is: `zangle = π/2 − elevation_apparent_norot` (matches the entry-angle convention in AlphaBeta.py and DynamicMassFit.py; `orbit.zc` would be worse since it is only set for orbits above the escape velocity), and the `getAtmDensity_vect` call on scalars (matches `calcKB`).

One optional thought for a future PR: Ceplecha (1988, Sec. 2f) prescribes a +0.15 offset for lower-sensitivity all-sky mirror cameras, and there is the IIIAi subgroup (IIIA with a ≥ 5 AU) — both could be added later if needed.

Issue #68, opened 2026-07-14
Bug in AggregateAndPlot causes failure to plot shower charts


markmac99:
There's a logic flaw in AggregateAndPlot at the point where it calculates `lam_err` and `bet_err` which causes a crash if the uncertainties are missing or invalid. 

The code makes use of a `traj` variable, to test if the uncertainties are present.  However, this variable has only inadvertently been set, and will always be the last element in the shower_trajs array (by virtue of the enumerate at line 957. So, its not a good way to test if every traj in the entire array has valid uncertainties.  Furthermore, this traj variable potentially masks or at least can be confused with the temporary traj variable used in the arrays.
``` python
        # Get the errors
        if traj.uncertainties is not None:
            lam_err = np.array([traj.uncertainties.L_g for traj in shower_trajs])
            bet_err = np.array([traj.uncertainties.B_g for traj in shower_trajs])
        else:
            lam_err = np.array([1 for traj in shower_trajs])
            bet_err = np.array([1 for traj in shower_trajs])
```

Consequently, if one or more of the members of shower_traj has blank or missing uncertainties, in particular if uncertainties.L_g is not set, the process will crash at this point.

``` python
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/ubuntu/src/WesternMeteorPyLib/wmpl/Trajectory/AggregateAndPlot.py", line 2413, in <module>
    generateShowerPlots(output_dir, traj_list, min_members=30, P_0m=P_0M, max_radiant_err=0.5)
  File "/home/ubuntu/src/WesternMeteorPyLib/wmpl/Trajectory/AggregateAndPlot.py", line 1018, in generateShowerPlots
    lam_err = np.array([traj.uncertainties.L_g for traj in shower_trajs])
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/src/WesternMeteorPyLib/wmpl/Trajectory/AggregateAndPlot.py", line 1018, in <listcomp>
    lam_err = np.array([traj.uncertainties.L_g for traj in shower_trajs])
                        ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'L_g'
```

I have a fix that I'm testing and will create a PR shortly.

PR #70, opened 2026-07-14
Fix for bug in AggregateAndPlot if some trajectories do not have uncertainties

markmac99:
See issue #68 for more info.

This corrects the logic used to create the arrays of uncertainty in lam and bet. Previously this would crash if one or more trajectories did not have an L_g member of the uncertainties.


markmac99:
I tested this with a variety of daily, monthly and annual datasets from UKMON, both with and without problematic trajectories.

PR #75, opened 2026-07-15
Alpha-Beta joint dynamics+light-curve fit + robust dynamics fit

AstroEloy:
- Add `fitAlphaBetaLightCurve()`: joint fit of the alpha-beta model to dynamics AND the
  light curve (Gritsevich & Koschny 2011), alongside the existing `fitAlphaBeta()`
  (dynamics-only) and `fitAlphaBetaMass()` (mass-constrained).
- Add a robust (soft-L1 least-squares) alternative to the classic Q4 dynamics fit, available
  both standalone (`fitAlphaBeta(method='robust')`) and inside the joint fit
  (`fitAlphaBetaLightCurve(dyn_method='robust')`).
- Add `alphaBetaLuminousEfficiency()` to recover the luminous efficiency tau from a fitted
  light-curve amplitude K.

New: joint dynamics + light-curve fit

`fitAlphaBetaLightCurve(v_data, ht_data, ht_lc_data, mag_abs_data, ...)` fits (alpha, beta)
jointly against the dynamics (height vs. velocity) and the light curve (height vs. absolute
magnitude), for a set of fixed mu values (default 0 and 2/3), using `scipy.optimize.least_squares`
with a soft-L1 robust loss. The amplitude (magnitude offset M0 / K) is profiled out analytically
at every iteration, so the nonlinear search stays 2-dimensional per mu.

Supporting additions:
- `alphaBetaHeightNormed()` / `alphaBetaVelocityNormed()`: normalized-space forward/inverse
  alpha-beta model. `alphaBetaHeight()`/`alphaBetaVelocity()` now delegate to these - the inverse
  now uses `brentq` instead of a per-point `scipy.optimize.minimize`, which is both faster and
  far more precise (round-trip error ~4e-9 m/s vs. the previous unconstrained-tolerance search).
- `alphaBetaLuminosityF()` / `alphaBetaModelMagnitude()`: the luminosity model (Eq. 13-14).
- `P_0M`, `ALPHA_BETA_BOUNDS` module constants (shared bounds/reference power used across fits).

Statistical robustness (informed by a domain-expert review of the initial design):
- `sigma_mag` accepts `None` (default - auto-derived per mu, from the MAD of an unweighted
  preliminary photometry-only fit at that mu, mirroring how `sigma_v` is derived from the Q4
  fit), a scalar, or a per-point array - the last is preferred whenever real photometric error
  estimates are available, since photometric uncertainty is strongly heteroscedastic.
- `dyn_weight` / `lc_weight`: optional multipliers to rebalance the dynamics vs. light-curve
  residual blocks when one has far more points than the other (off by default).
- Docstring explicitly flags the statistical caveats: `sigma_v` is an *effective* uncertainty
  (absorbs model misspecification, not pure instrument noise); `alpha_std_rel`/`beta_std_rel`
  are local Gauss-Newton curvature estimates under the robustified soft-L1 cost, not rigorous
  covariances; residuals are treated as independent (no inter-point correlation modeling);
  `cost` (and anything derived from it, like `best_fixed_mu`) is only comparable across mu when
  `sigma_mag` is not auto-derived - a runtime `NOTE` fires whenever this could matter.
- `fit_free_mu=True` (opt-in): fits mu itself as a 3rd free parameter (bounded [0, 2/3]) instead
  of only comparing it at the fixed `mu_values` grid - returned as `mu_free_fit`, drawn as a
  distinct curve when `plot=True`. `best_fixed_mu` (renamed from `best_mu`) is only an argmin
  over the fixed grid, not a continuous optimum - `mu_free_fit` is.
- `plot=True` draws a two-panel plot (velocity, magnitude) with joint / dynamics-only /
  photometry-only curves per mu, each legend entry self-descriptive and one line; axis limits are
  clipped to the observed data range so a poorly-constrained diagnostic curve can't blow out the
  view.
- Both `fitAlphaBeta()`/`fitAlphaBetaMass()`/`fitAlphaBetaLightCurve()` now document explicitly
  that `ht_data`/`ht_lc_data` must already be rescaled to the equivalent exponential atmosphere
  (`rescaleHeightToExponentialAtmosphere()`) - an existing precondition that was previously only
  implicit in the `__main__` CLI usage.

`alphaBetaLuminousEfficiency(K, alpha, beta, slope, v_init, mu=0, dens=..., shape_coeff=...,
gamma=..., v_final=None)` recovers the dimensionless luminous efficiency tau from a fitted K
(`tau = K*2*HT_NORM_CONST/(M_e*V_e^3*sin(slope))`), reusing `alphaBetaMasses()` for M_e - a
deliberately separate step, since the light-curve fit itself makes no density/shape/drag
assumption.

New: robust dynamics-only fit

The classic Q4 fit (`minimizeAlphaBeta()`, Gritsevich 2007) does not minimize the velocity
residual `v_model - v_obs` directly - its residual is a transformed quantity whose
`exp(-ht_normed)` factor structurally under-weights the end of the trajectory (where the
deceleration/beta signal actually is) relative to the near-constant-velocity beginning. On
synthetic data, this measurably hurts the fit specifically where it matters: RMSE near the
trajectory's end (bottom 20% height) was ~806 m/s for Q4 vs. ~297 m/s for the new method.

- `minimizeAlphaBetaRobust()`: least squares directly on the velocity residuals, robustified
  with soft-L1 by default, seeded from the Q4 result.
- `fitAlphaBeta(..., method='q4'|'robust', sigma_v=None, loss='soft_l1', f_scale=2.0)`:
  `method='q4'` (default) is byte-for-byte unchanged; `method='robust'` uses the new solver.
- `fitAlphaBetaLightCurve(..., dyn_method='q4'|'robust')`: lets the Stage-0 dynamics-only
  reference (and joint-fit seed) use the robust method too. `sigma_v` is always derived from the
  Q4 fit's residuals specifically (even when `dyn_method='robust'`), to avoid the circularity of
  a fit weighting itself by its own residuals.
- Extracted `_logAlphaBetaBounds()`, `_estimateSigmaV()`, `_dynResiduals()` to remove duplication
  between `fitAlphaBeta()` and `fitAlphaBetaLightCurve()`.

Other

- `minimizeAlphaBeta()` no longer prints an unconditional `"Initial guess:"` debug line.
- `_makeBoundChecker()` factory extracted and reused by `fitAlphaBetaMass()` (previously an
  inline closure duplicated per call site) and `fitAlphaBetaLightCurve()`.
- Compacted several `scipy.optimize.least_squares(...)` call sites to match the single-line/
  minimal-continuation style already used for `scipy.optimize.minimize(...)` elsewhere in this
  module, and extracted a shared `lsq_kwargs` dict to remove 5x-repeated boilerplate.

Compatibility

All new parameters are additive and default to the prior behavior (`method='q4'`,
`dyn_method='q4'`, `fit_free_mu=False`, `dyn_weight=lc_weight=1.0`); `fitAlphaBeta()`'s and
`fitAlphaBetaMass()`'s return signatures are unchanged. `fitAlphaBetaLightCurve()` is a new
function, so it has no prior callers to break.

PR hilghly assisted by Claude

### DIFF
--- a/wmpl/Utils/AlphaBeta.py
+++ b/wmpl/Utils/AlphaBeta.py
@@ -341,7 +341,7 @@ def _betaFromMasses(mu, mass_initial, mass_final, v_final, v_init):
     return (1.0 - mu)*np.log(mass_initial/mass_final)/(1.0 - (v_final/v_init)**2)
 
 
-def _makeBoundChecker(warnings_list, tol=1e-2):
+def _makeBoundChecker(warnings_list, tol=1e-2, log=False):
     """ Build a _checkBound(name, value, lower, upper) closure that appends a message to
         warnings_list whenever value lands within a relative tol of lower or upper. Shared by
         fitAlphaBetaMass() and fitAlphaBetaLightCurve() to flag fitted parameters that are
@@ -352,6 +352,12 @@ def _makeBoundChecker(warnings_list, tol=1e-2):
 
     Keyword arguments:
         tol: [float] Relative distance-to-bound tolerance.
+        log: [bool] If True, measure the distance to the bounds in log space instead of linear
+            space. Appropriate for parameters whose bounds span several orders of magnitude
+            (alpha, beta): a linear check would flag almost any physically reasonable value as
+            "close to the lower bound" (e.g. beta=0.33 is only 0.7% of the way across the linear
+            (1e-5, 50) span, but sits comfortably in the middle of it in log space). Requires
+            value, lower and upper to be positive.
 
     Return:
         _checkBound: [callable] (name, value, lower, upper) -> None.
@@ -359,14 +365,19 @@ def _makeBoundChecker(warnings_list, tol=1e-2):
 
     def _checkBound(name, value, lower, upper):
 
-        span = upper - lower
+        if log:
+            check_value, check_lower, check_upper = np.log(value), np.log(lower), np.log(upper)
+        else:
+            check_value, check_lower, check_upper = value, lower, upper
 
-        if (value - lower)/span < tol:
+        span = check_upper - check_lower
+
+        if (check_value - check_lower)/span < tol:
             warnings_list.append(
                 "{:s}={:.6g} is close to the fit lower bound ({:.6g})".format(name, value, lower)
             )
 
-        if (upper - value)/span < tol:
+        if (check_upper - check_value)/span < tol:
             warnings_list.append(
                 "{:s}={:.6g} is close to the fit upper bound ({:.6g})".format(name, value, upper)
             )
@@ -1477,6 +1488,11 @@ def fitAlphaBetaMass(
 
     _checkBound = _makeBoundChecker(warnings_list, tol=bound_tol)
 
+    # Beta spans several orders of magnitude between its bounds, so its distance to them is only
+    #   meaningful in log space (a linear check would flag e.g. beta=0.33 as "close" to the 1e-5
+    #   lower bound) - see _makeBoundChecker()
+    _checkBoundLog = _makeBoundChecker(warnings_list, tol=bound_tol, log=True)
+
     # Check the fitted densities against the optimization bounds
     if fit_density:
         _checkBound("density_mu0", density_mu0, xmin[0], xmax[0])
@@ -1484,14 +1500,14 @@ def fitAlphaBetaMass(
 
     # Check the fitted betas (under "both" beta is analytic, not fitted, so it's skipped)
     if mass_constraint in ["initial", "final"]:
-        _checkBound("beta_mu0", beta_mu0, xmin[1], xmax[1])
-        _checkBound("beta_mu23", beta_mu23, xmin[1], xmax[1])
+        _checkBoundLog("beta_mu0", beta_mu0, xmin[1], xmax[1])
+        _checkBoundLog("beta_mu23", beta_mu23, xmin[1], xmax[1])
 
     if fit_density and mass_constraint in ["final", "both"]:
         _checkBound("density_mu_best", density_mu_best, xmin[0], xmax[0])
 
     if mass_constraint == "final":
-        _checkBound("beta_mu_best", beta_mu_best, xmin[1], xmax[1])
+        _checkBoundLog("beta_mu_best", beta_mu_best, xmin[1], xmax[1])
 
     def _checkVelocityRecovery(name, alpha, beta):
         """ Under mass_constraint="final"/"both", beta is derived analytically assuming a
@@ -1888,6 +1904,15 @@ def fitAlphaBetaLightCurve(
         light curve (height vs. absolute magnitude), following Gritsevich (2007, 2009) for the
         dynamics and Gritsevich & Koschny (2011) for the luminosity.
 
+    WARNING: Like the whole alpha-beta formalism, this fit assumes a SINGLE, non-fragmenting
+        body ablating quasi-continuously. Most real fireballs fragment, and a fragmenting event's
+        light curve (flares, sudden brightenings) does not follow Eq. (13)-(14) below - so this
+        joint fit, and especially the photometric outputs derived from it (M0, K, and any
+        luminous efficiency computed with alphaBetaLuminousEfficiency()), is NOT expected to be
+        valid for the majority of real events. Only apply it to events consistent with
+        single-body ablation, and inspect the light curve fit visually (plot=True) before
+        trusting the photometric results.
+
     Dynamics, dimensionless v = V/V_e, y = h/HT_NORM_CONST (see alphaBetaHeightNormed()):
 
         y = ln(2 alpha) + beta - ln(Delta),   Delta = Ei(beta) - Ei(beta v^2)          (7)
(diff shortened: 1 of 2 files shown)


dvida:
Thanks @AstroEloy, this is a solid contribution — the math core checks out and the statistical caveats are documented honestly. I did a detailed review and pushed 3 commits directly to your branch (your fork's `master`, so please `git pull` before making any further changes):

- **063b9fc — Fix profiled-amplitude weighting and minor robustness gaps.** The main finding: `_profiledMagOffset()` used an unweighted median even when a per-point `sigma_mag` was given, so noisy faint-end points pulled M0 as hard as precise peak points — biasing K and any derived luminous efficiency exactly in the recommended per-point-sigma usage. It now computes the 1/sigma-weighted median (reduces exactly to the plain median for scalar/uniform sigma, so results without per-point sigmas are unchanged). In a Monte-Carlo check with skewed faint-end noise, the M0 bias went from −0.085 mag (≈8% error in K) to −0.003 mag. Smaller items in the same commit: `best_fixed_mu` now prefers converged fits over failed ones; the log-space seed is clipped in linear space first (a non-positive Q4 seed could produce a NaN `x0` on older SciPy); `mu_values` are validated up front; docstrings no longer claim the profiled M0 is the exact conditional optimum of the soft-L1 cost (it is exact for the sigma-normalized L1 cost), and the input-ordering convention for the derived `v_init` is documented.
- **9246415 — Tests.** 7 new tests in `wmpl/Utils/Tests/test_AlphaBeta.py` covering the normed model round trip, `fitAlphaBeta(method='robust')`, the luminosity model, the weighted profiled offset (regression guard), `fitAlphaBetaLightCurve()` across all three `sigma_mag` modes plus the free-mu fit, the luminous-efficiency inversion, and input validation. All 13 tests (6 existing + 7 new) pass.
- **06f5c1e — Style.** Structured `Arguments:`/`Return:` docstring sections on the new private helpers, and small formatting consistency fixes.

What I verified as correct so you don't need to worry about it: the `alphaBetaHeight()`/`alphaBetaVelocity()` delegation is algebraically identical to the old code; the `brentq` bracket cannot fail (dy/dv > 0 on (0,1)); no overflow/log-of-negative in the luminosity path for mu in [0, 2/3]; the magnitude/K convention is self-consistent with Eq. (13); and the tau inversion is dimensionally correct.

Before I merge, could you confirm:

1. **The weighted-median change** — do you agree with weighting the profiled M0 by 1/sigma_mag? It changes results whenever a per-point `sigma_mag` array is passed (scalar/None behavior is identical to your original).
2. **Eq. (14) transcription** — the one thing I couldn't independently close: could you double-check the implemented `f(v)` in `alphaBetaLuminosityF()` against the actual Gritsevich & Koschny (2011) paper text? The code matches its own docstring and is internally consistent with Eq. (13), but a cross-check against the paper would be good.
3. **The `mu_values` validation** — values outside [0, 1) now raise a `ValueError` up front (with a warning above 2/3). Any workflow of yours that relied on passing other values?
4. Please run `python -m wmpl.Utils.Tests.test_AlphaBeta` (or pytest) on your side and confirm everything passes with your setup and real-data use cases.

If all four look good to you, I'll merge.

AstroEloy:
Thanks @dvida for the improvements.

1. Looks good to me. The weighted median is the consistent choice when per-point sigma_mag values are available, while the scalar/None behavior remains unchanged.
2. I have doubled checked the equation with the original paper. Everything is ok.
3. Not really, only 0 to 2/3.
4. All 13 tests passed + some additional test with real-data.

AstroEloy:
I just saw this in the test output that need to be fixed:

- beta_mu23=0.329999 is close to the fit lower bound (0.001)

AstroEloy:
I recommend adding a clear warning to the fitAlphaBetaLightCurve() docstring stating that the method is only applicable to single-bodies. Since most fireballs undergo fragmentation, the fit is not expected to be valid for the majority of real events.

dvida:
Thanks for the quick confirmation and the paper cross-check @AstroEloy. Both of your follow-up points are addressed in 3969e1c:

- **Spurious bound warning fixed.** `fitAlphaBetaMass()` was checking beta's distance to its optimization bounds in linear space; since beta's bounds span several orders of magnitude (1e-5 to 50), values like `beta_mu23=0.33` got flagged as "close to the lower bound". `_makeBoundChecker()` now supports log-space checking, which is used for all alpha/beta bound checks (and also replaces the duplicated inline log-space checker in `fitAlphaBetaLightCurve()`). The warning no longer fires in the test output. Note that the `v_final_mu23` self-consistency warnings you may still see when running the tests are genuine diagnostics (the mu=2/3 branch fitted to mu=0-generated synthetic data), not false positives.
- **Single-body warning added.** `fitAlphaBetaLightCurve()`'s docstring now opens with a prominent WARNING that the fit assumes a single, non-fragmenting body and is therefore not expected to be valid for the majority of real (fragmenting) fireballs, with a pointer to inspect the light curve fit visually (`plot=True`) before trusting M0/K/tau.

All 13 tests pass. From my side this is now ready to merge.

AstroEloy:
Thanks! I think is ready to go.

PR #76, opened 2026-07-16
Add analytic error estimation to fitAlphaBeta(method='robust')

AstroEloy:
## Summary

Adds analytic uncertainty estimation to `fitAlphaBeta()`, previously a bare `(v_init, alpha, beta)`
point estimate with no error information at all (unlike `fitAlphaBetaLightCurve()`, which already
carries an approximate Gauss-Newton covariance for exactly this reason).

- `estimate_errors=True` (new kwarg, `method='robust'` only — Q4's L1/Nelder-Mead fit has no
  Jacobian to build a covariance from, and raises `ValueError` if requested there) returns a 4th
  element, an `errors` dict, alongside the unchanged `(v_init, alpha, beta)`. Fully backward
  compatible: default `estimate_errors=False` leaves every existing call site untouched.
- Covariance of `(ln alpha, ln beta)` from the fit's own Gauss-Newton `(J^T J)^-1` (reusing the
  Jacobian `minimizeAlphaBetaRobust()` already computes internally, via a new
  `return_fit_result=True` kwarg there, instead of recomputing it), reduced-chi-square-scaled —
  same recipe `fitAlphaBetaLightCurve()` already uses for its own `alpha_std_rel`/`beta_std_rel`.
- `sigma_v_init` (new kwarg): propagates `v_init`'s own uncertainty into alpha/beta's
  covariance via a numerically estimated sensitivity `d(ln alpha, ln beta)/d(v_init)` (2 extra
  refits at `v_init +/- a small step`), added as a rank-1 correction. If `v_init` is derived
  internally (median of the leading ~20%/min-10 points) rather than supplied, its uncertainty is
  auto-derived as the **standard error of that median** (`1.2533*1.4826*MAD/sqrt(N)`), not the
  plain MAD.
- Both log-space (`cov_log`, `corr_log` — the space this model is actually fit in) and linear-space
  (`cov`, `corr`, `alpha_std`, `beta_std`, via the exact delta-method transform
  `cov = diag(alpha,beta) @ cov_log @ diag(alpha,beta)`) are returned, so a caller can propagate
  into any derived quantity `f(alpha, beta)` via `sigma_f^2 = J cov J^T` without re-deriving the
  transform themselves.
- `ci` (new kwarg, default 95.0) plus `alpha_ci_gaussian_lower/upper`,
  `beta_ci_gaussian_lower/upper`: a parametric, log-normal confidence interval. Named
  `..._gaussian_...` deliberately — WMPL uses `confidenceInterval()`/`_95ci_lower/upper` elsewhere
  (`Trajectory.py`, `REBOUND.py`, `FitPopulationAndMassIndex.py`) for **empirical** Monte Carlo
  percentiles; there is no ensemble here, only a Gaussian-in-log-space approximation, and the name
  is meant to make that impossible to miss.
- `verbose` (new kwarg): prints a summary (v_init/alpha/beta, and the error estimate if
  requested), matching `fitAlphaBetaLightCurve()`'s existing print style.

Correctness fix included: a fit landing at/near an `ALPHA_BETA_BOUNDS` edge (optimizer clipped
against the box constraint rather than converged to an interior optimum) can produce a
near-singular Jacobian whose `(J^T J)^-1` comes back with a negative variance on the diagonal —
mathematically impossible for a real covariance, and not caught by `except LinAlgError` (a
near-singular matrix doesn't raise). This previously surfaced as silent NaN plus a bare
`RuntimeWarning: invalid value encountered in sqrt` from deep inside `sqrt()`, with no indication
of the actual cause. Now: a log-space bound-proximity check (same convention
`fitAlphaBetaLightCurve()` already uses) warns explicitly whenever alpha/beta lands within 1% of a
bound, independent of whether the covariance itself is still numerically usable; and an explicit
negative-diagonal guard returns clean NaN with a clear message before any `sqrt()` call is
attempted. Found via a real trajectory in the accompanying notebook (an unrealistic `v_init`
pinned beta at 49.99995, a hair under its upper bound of 50).

## Test plan

- [x] `wmpl/Utils/Tests/test_AlphaBeta.py`: 7 new tests - `method='q4'` rejection, point-estimate
      invariance under `estimate_errors`, covariance structure and the exact log<->linear
      delta-method identities, confidence-interval bracketing/widening, `verbose=True`,
      `sigma_v_init` propagation (both explicit and auto-derived), and the bound-pinned failure
      mode (regression test using a synthetic true beta of 150 - well beyond the bound of 50 -
      which reproduces the exact same pinned value, 49.999950, seen on real data)
- [x] All 20 tests in the file pass (`python -m wmpl.Utils.Tests.test_AlphaBeta`)
- [x] Verified against the real fireballs

### DIFF
--- a/wmpl/Utils/AlphaBeta.py
+++ b/wmpl/Utils/AlphaBeta.py
@@ -642,7 +642,10 @@ def _alphaBetaRobustErrors(fit_res, alpha, beta, v_data, ht_data, v_init, v_init
 
         # Standard error of the median (NOT the plain MAD, which measures the dispersion of the
         #   observations themselves, not the precision of the median as an estimator of v_init)
-        #   of the same leading points fitAlphaBeta() used to derive v_init.
+        #   of the same leading points fitAlphaBeta() used to derive v_init. This captures the
+        #   RANDOM error of that estimator, but not its systematic one: the median of a still-
+        #   decelerating leading window is itself biased low relative to the true v_init (see
+        #   fitAlphaBeta()'s docstring caveat on this - no variance term can correct for a bias).
         max_index = max(int(0.2*len(v_data)), 10)
         v_head = v_data[:max_index]
         mad = np.median(np.abs(v_head - np.median(v_head)))
@@ -825,7 +828,11 @@ def fitAlphaBeta(v_data, ht_data, v_init=None, method='q4', sigma_v=None, loss='
                 if it was given a value) - v_init's uncertainty is instead estimated automatically
                 as the standard error of the median of the same leading points used to derive it
                 (see the Return section), which is the quantity that actually needs to be
-                propagated, not the raw scatter of those points.
+                propagated, not the raw scatter of those points. IMPORTANT: this covers only the
+                RANDOM error of that median - it is also systematically biased low by the
+                deceleration trend across the leading window itself (a fixed offset, not
+                something a variance term can correct for), which can meaningfully under-cover
+                alpha_ci_gaussian_lower/upper - see the Statistical caveats below.
         ci: [float] Confidence level (0-100, e.g. 95, the default) for the approximate
             alpha_ci_gaussian_lower/upper and beta_ci_gaussian_lower/upper bounds below - only
             used if estimate_errors=True. Same convention as
@@ -896,6 +903,22 @@ def fitAlphaBeta(v_data, ht_data, v_init=None, method='q4', sigma_v=None, loss='
           is NaN (alpha/beta themselves are still returned). Both mean the same thing: the fit is
           likely poorly constrained by this data, and the local curvature approximation isn't
           meaningful around a boundary the optimizer was clipped against rather than converged to.
+        - With v_init=None (derived internally), alpha_ci_gaussian_lower/upper can under-cover the
+          true alpha/beta by a lot more than the other approximations above: the leading window
+          used to derive v_init is still decelerating, so its median is not just noisy but
+          systematically biased LOW relative to the true v_init (-11 m/s, deterministic, on this
+          module's own synthetic test geometry - a fixed offset, not something sigma_v_init's
+          variance term can correct for). Measured on that same geometry (300-realization Monte
+          Carlo): at the noise level used elsewhere in this test suite, the nominal 95%
+          alpha_ci_gaussian_lower/upper actually covers the true alpha/beta only ~65-71% of the
+          time with a derived v_init - even though alpha_std_rel/beta_std_rel themselves stay
+          accurate - recovering to ~90% at 3x that noise as the fixed bias shrinks relative to the
+          growing scatter. With v_init supplied externally instead (no derivation, hence no
+          bias), coverage is a nominal ~96%. The bias is in the pre-existing v_init derivation,
+          not introduced by estimate_errors - but the confidence interval is what turns a
+          previously-inconsequential offset into a miscalibrated claim. Prefer an external v_init
+          (+ sigma_v_init if it has its own uncertainty) over the internally-derived one whenever
+          interval coverage matters, not just the point estimate.
     """
 
     method = method.lower()
--- a/wmpl/Utils/AlphaBeta.py
+++ b/wmpl/Utils/AlphaBeta.py
@@ -479,7 +479,7 @@ def _dynResiduals(x, v_normed, ht_normed, sigma_v_normed):
 
 
 def minimizeAlphaBetaRobust(v_normed, ht_normed, sigma_v_normed, loss='soft_l1', f_scale=2.0,
-        x0_alpha_beta=None):
+        x0_alpha_beta=None, return_fit_result=False):
     """ Robust alternative to minimizeAlphaBeta(): least squares (soft-L1 by default) directly on
         the velocity residuals (v_model - v_obs), instead of Q4's transformed residual.
 
@@ -506,9 +506,13 @@ def minimizeAlphaBetaRobust(v_normed, ht_normed, sigma_v_normed, loss='soft_l1',
         f_scale: [float] Robust loss scale, in units of sigma_v_normed.
         x0_alpha_beta: [tuple] Optional (alpha0, beta0) initial guess. If None, derived from
             minimizeAlphaBeta().
+        return_fit_result: [bool] If True, also return the raw scipy.optimize.least_squares
+            result object as a 3rd element, so a caller can reuse its Jacobian/cost for a
+            Gauss-Newton covariance (see fitAlphaBeta()'s estimate_errors) instead of re-fitting
+            or recomputing the Jacobian from scratch.
 
     Return:
-        (alpha, beta): [tuple of floats].
+        (alpha, beta), or (alpha, beta, res) if return_fit_result=True.
     """
 
     if x0_alpha_beta is None:
@@ -526,11 +530,235 @@ def minimizeAlphaBetaRobust(v_normed, ht_normed, sigma_v_normed, loss='soft_l1',
     if not res.success:
         print("WARNING: Optimizer failed: {:s}".format(res.message))
 
+    if return_fit_result:
+        return tuple(np.exp(res.x)) + (res,)
+
     return tuple(np.exp(res.x))
 
 
+def _alphaBetaRobustErrors(fit_res, alpha, beta, v_data, ht_data, v_init, v_init_given, sigma_v,
+        sigma_v_init, loss, f_scale, ci):
+    """ Analytic error estimate for fitAlphaBeta(method='robust', estimate_errors=True).
+
+        Two contributions to the covariance of (ln alpha, ln beta):
+        1. The Gauss-Newton local curvature of the robust-loss least_squares fit itself
+           (fit_res.jac/.cost/.fun, reused from minimizeAlphaBetaRobust() rather than
+           recomputed), treating v_init as if it were exactly known. Same (J^T J)^-1,
+           reduced-chi-square-scaled recipe as fitAlphaBetaLightCurve()'s own
+           alpha_std_rel/beta_std_rel, with the same caveat: soft-L1 reweights residuals
+           nonlinearly, so this is a Gauss-Newton approximation of the local curvature, not the
+           inverse Fisher information of Gaussian noise.
+        2. A rank-1 correction J_v0 sigma_v_init^2 J_v0^T propagating v_init's OWN uncertainty
+           through a numerically estimated sensitivity J_v0 = d(ln alpha, ln beta)/d(v_init) -
+           obtained by refitting fitAlphaBeta() itself at v_init +/- a small step, since v_init is
+           held fixed during the fit that produced fit_res and so never appears in fit_res.jac.
+           Skipped (zero contribution) whenever sigma_v_init is 0 (v_init treated as exact).
+        The two contributions are simply added (v_init's estimation error and the conditional
+        fit's residuals are treated as independent, even though both ultimately come from the
+        same v_data) - a standard plug-in approximation, not a fully joint treatment.
+
+        cov_log/corr_log ((ln alpha, ln beta), the space this model is actually fit in) are
+        propagated to linear (alpha, beta) space via the delta method: since
+        d(alpha)/d(ln alpha) = alpha, the Jacobian of (alpha, beta) w.r.t. (ln alpha, ln beta) is
+        diag(alpha, beta), so cov = diag(alpha, beta) @ cov_log @ diag(alpha, beta). corr comes
+        out numerically identical to corr_log (correlation is invariant under this kind of
+        positive-diagonal rescaling) - both are kept for convenience.
+
+        alpha_ci_gaussian_lower/upper and beta_ci_gaussian_lower/upper are APPROXIMATE confidence
+        intervals based on the local Gaussian approximation in log-space (equivalently: assuming
+        alpha/beta are log-normal), at the `ci` level - exp(ln(alpha) +/- z*alpha_std_rel),
+        asymmetric around alpha and, unlike alpha +/- z*alpha_std, never negative. The "_gaussian"
+        in the name (and "approximate" here) is deliberate: this is NOT the same kind of object as
+        wmpl.Utils.Math.confidenceInterval()'s output elsewhere in this library (e.g. Trajectory.py,
+        REBOUND.py) - those are empirical percentiles of an actual Monte Carlo ensemble, whereas
+        there is no ensemble here, only a parametric approximation. Don't drop the "_gaussian"
+        qualifier when reading these out; it is the one thing standing between this and a false
+        claim of MC-equivalent rigor.
+
+        Two failure modes are handled explicitly, both printing a WARNING and returning NaN for
+        every error field (alpha/beta themselves are still returned - only their errors are NaN):
+        - (alpha, beta) landed within 1% (log-space) of an ALPHA_BETA_BOUNDS edge: the fit is
+          likely poorly constrained by this data, and the local curvature below isn't meaningful
+          around a boundary the optimizer was clipped against rather than converged to.
+        - The Jacobian is near- (but not exactly) singular: (J^T J)^-1 comes out with a negative
+          diagonal - impossible for a real covariance, and NOT the same thing as the outright
+          LinAlgError case below (a near-singular matrix does not raise). This is what a
+          bound-pinned fit typically causes, but is checked independently since it is the actual
+          numerical symptom regardless of cause.
+
+    Arguments:
+        fit_res: [OptimizeResult] From minimizeAlphaBetaRobust(..., return_fit_result=True).
+        alpha: [float] Fitted ballistic coefficient.
+        beta: [float] Fitted mass loss parameter.
+        v_data: [ndarray] Velocity data (m/s), as given to fitAlphaBeta().
+        ht_data: [ndarray] Height data (m), as given to fitAlphaBeta().
+        v_init: [float] Adopted initial velocity (m/s).
+        v_init_given: [bool] Whether v_init was supplied by the caller (vs. derived internally).
+        sigma_v: [float] Velocity uncertainty (m/s) used to weight the fit.
+        sigma_v_init: [float or None] Caller-supplied v_init uncertainty - only used if
+            v_init_given (see fitAlphaBeta()'s docstring).
+        loss: [str] As in fitAlphaBeta().
+        f_scale: [float] As in fitAlphaBeta().
+        ci: [float] Confidence level (0-100, e.g. 95) for alpha_ci_gaussian_lower/upper and
+            beta_ci_gaussian_lower/upper - same convention as
+            wmpl.Utils.Math.confidenceInterval()'s `ci`.
+
+    Return:
+        errors: [dict] 'alpha', 'beta', 'alpha_std_rel', 'beta_std_rel', 'alpha_std', 'beta_std',
+            'cov_log', 'corr_log', 'cov', 'corr', 'alpha_ci_gaussian_lower',
+            'alpha_ci_gaussian_upper', 'beta_ci_gaussian_lower', 'beta_ci_gaussian_upper', 'ci',
+            'sigma_v', 'sigma_v_init' - see fitAlphaBeta()'s docstring for details.
+    """
+
+    # Resolve v_init's own uncertainty up front, so it is available in `errors` even if the
+    #   covariance itself turns out to be degenerate below (see the bound/near-singular checks).
+    if v_init_given:
+
+        if sigma_v_init is None:
+            print("WARNING: v_init was given without sigma_v_init - its uncertainty will NOT be "
+                "propagated into alpha_std_rel/beta_std_rel (v_init is being treated as exactly "
+                "known).")
+            sigma_v_init_used = 0.0
+
+        else:
+            sigma_v_init_used = sigma_v_init
+
+    else:
+        # Standard error of the median (NOT the plain MAD, which measures the dispersion of the
+        #   observations themselves, not the precision of the median as an estimator of v_init)
+        #   of the same leading points fitAlphaBeta() used to derive v_init.
+        max_index = max(int(0.2*len(v_data)), 10)
+        v_head = v_data[:max_index]
+        mad = np.median(np.abs(v_head - np.median(v_head)))
+        sigma_v_init_used = 1.2533*1.4826*mad/np.sqrt(max_index)
+
+    nan_result = {
+        'alpha': alpha, 'beta': beta, 'alpha_std_rel': np.nan, 'beta_std_rel': np.nan,
+        'alpha_std': np.nan, 'beta_std': np.nan, 'cov_log': np.full((2, 2), np.nan),
+        'corr_log': np.nan, 'cov': np.full((2, 2), np.nan), 'corr': np.nan,
+        'alpha_ci_gaussian_lower': np.nan, 'alpha_ci_gaussian_upper': np.nan,
+        'beta_ci_gaussian_lower': np.nan, 'beta_ci_gaussian_upper': np.nan,
+        'ci': ci, 'sigma_v': sigma_v, 'sigma_v_init': sigma_v_init_used,
+    }
+
+    # Warn if the fitted (alpha, beta) landed near an optimization bound, using the same
+    #   log-space tolerance fitAlphaBetaLightCurve() uses for its own bound check (alpha/beta
+    #   span orders of magnitude, so a linear-space check would misfire - see
+    #   _makeBoundChecker()). The Gauss-Newton curvature below is only meaningful around a
+    #   genuine local optimum, and tends to be degenerate (or numerically nonsensical) once the
+    #   fit has been clipped against a box constraint instead of converging to one.
+    log_lower, log_upper = _logAlphaBetaBounds()
+    bound_tol = 1e-2
+    bound_warnings = []
+
+    for name, value, lo, hi in [
+            ("alpha", alpha, log_lower[0], log_upper[0]),
+            ("beta", beta, log_lower[1], log_upper[1])]:
+
+        log_value = np.log(value)
+        span = hi - lo
+
+        if (log_value - lo)/span < bound_tol:
+            bound_warnings.append("{:s}={:.6g} is close to the fit lower bound ({:.6g})".format(
+                name, value, np.exp(lo)))
+
+        if (hi - log_value)/span < bound_tol:
+            bound_warnings.append("{:s}={:.6g} is close to the fit upper bound ({:.6g})".format(
+                name, value, np.exp(hi)))
+
+    if bound_warnings:
+        print()
+        print("WARNING: fitted (alpha, beta) landed within {:.3g}% of an optimization bound - "
+            "the fit may be poorly constrained, and the Gauss-Newton error estimate below may "
+            "be degenerate (NaN) or unreliable as a result.".format(100*bound_tol))
+        for warning in bound_warnings:
+            print("  - {:s}".format(warning))
+        print()
+
+    # Gauss-Newton covariance of (ln alpha, ln beta), v_init held fixed at its adopted value
+    dof = max(len(fit_res.fun) - len(fit_res.x), 1)
+    s2 = 2.0*fit_res.cost/dof
+
+    try:
+        cov_fit = s2*np.linalg.inv(fit_res.jac.T @ fit_res.jac)
+    except np.linalg.LinAlgError:
+        return nan_result
+
+    # A near- (but not exactly) singular Jacobian can produce a numerically "valid" (no
+    #   LinAlgError) inverse with a negative variance on the diagonal - mathematically impossible
+    #   for a real covariance, and NOT caught by the try/except above. Most often seen exactly
+    #   when (alpha, beta) is pinned against a bound (see the warning above): fail cleanly with
+    #   NaN here instead of letting np.sqrt() silently emit a generic RuntimeWarning below.
+    if np.any(np.diag(cov_fit) < 0):
+        print("WARNING: the Gauss-Newton covariance has a negative variance on its diagonal "
+            "(a near-singular Jacobian, most often from a fit pinned against a bound - see "
+            "above). Returning NaN for alpha_std_rel/beta_std_rel/cov/corr/the confidence "
+            "interval instead of a meaningless value.")
+        return nan_result
+
+    # Rank-1 correction from v_init's uncertainty, via a numerical sensitivity. Skipped if there
+    #   is no v_init uncertainty to propagate.
+    if sigma_v_init_used > 0:
+
+        dv = 1e-4*v_init
+
+        def _logAlphaBetaAt(v_init_pert):
+            _, alpha_p, beta_p = fitAlphaBeta(v_data, ht_data, v_init=v_init_pert, \
+                method='robust', sigma_v=sigma_v, loss=loss, f_scale=f_scale)
+            return np.log([alpha_p, beta_p])
+
+        jac_v0 = (_logAlphaBetaAt(v_init + dv) - _logAlphaBetaAt(v_init - dv))/(2*dv)
+
+        cov_v0 = np.outer(jac_v0, jac_v0)*sigma_v_init_used**2
+
+    else:
+        cov_v0 = np.zeros((2, 2))
+
+    # cov_v0 is PSD (an outer product scaled by a non-negative variance), so it cannot turn the
+    #   already-checked non-negative diagonal of cov_fit negative - cov_log's diagonal is
+    #   guaranteed non-negative here too.
+    cov_log = cov_fit + cov_v0
+
+    alpha_std_rel, beta_std_rel = np.sqrt(np.diag(cov_log))
+    corr_log = cov_log[0, 1]/(alpha_std_rel*beta_std_rel)
+
+    # Linear (alpha, beta) space, via the delta method (see the docstring above)
+    scale = np.array([alpha, beta])
+    cov = np.outer(scale, scale)*cov_log
+    alpha_std, beta_std = np.sqrt(np.diag(cov))
+    corr = cov[0, 1]/(alpha_std*beta_std)
+
+    # Parametric (log-normal) confidence interval - see the docstring above for how this differs
+    #   from wmpl.Utils.Math.confidenceInterval()'s empirical percentiles elsewhere in the library
+    z = scipy.special.ndtri(0.5 + ci/200.0)
+    alpha_ci_gaussian_lower = alpha*np.exp(-z*alpha_std_rel)
+    alpha_ci_gaussian_upper = alpha*np.exp(z*alpha_std_rel)
+    beta_ci_gaussian_lower = beta*np.exp(-z*beta_std_rel)
+    beta_ci_gaussian_upper = beta*np.exp(z*beta_std_rel)
+
+    return {
+        'alpha': alpha,
+        'beta': beta,
+        'alpha_std_rel': alpha_std_rel,
+        'beta_std_rel': beta_std_rel,
+        'alpha_std': alpha_std,
+        'beta_std': beta_std,
+        'cov_log': cov_log,
+        'corr_log': corr_log,
+        'cov': cov,
+        'corr': corr,
+        'alpha_ci_gaussian_lower': alpha_ci_gaussian_lower,
+        'alpha_ci_gaussian_upper': alpha_ci_gaussian_upper,
+        'beta_ci_gaussian_lower': beta_ci_gaussian_lower,
+        'beta_ci_gaussian_upper': beta_ci_gaussian_upper,
+        'ci': ci,
+        'sigma_v': sigma_v,
+        'sigma_v_init': sigma_v_init_used,
+    }
+
+
 def fitAlphaBeta(v_data, ht_data, v_init=None, method='q4', sigma_v=None, loss='soft_l1',
-        f_scale=2.0):
+        f_scale=2.0, estimate_errors=False, sigma_v_init=None, ci=95.0, verbose=False):
     """ Fit the alpha and beta parameters to the given velocity and height data.
 
     Two methods are available (see minimizeAlphaBeta()/minimizeAlphaBetaRobust() for the full
@@ -543,7 +771,9 @@ def fitAlphaBeta(v_data, ht_data, v_init=None, method='q4', sigma_v=None, loss='
         - method='robust': least squares directly on the velocity residuals, robustified with
           `loss` (soft-L1 by default), seeded from the Q4 result. Weights the whole trajectory
           much more evenly than Q4 and tracks the deceleration/"knee" more faithfully - this is
-          the same approach fitAlphaBetaLightCurve() uses for its dynamics block.
+          the same approach fitAlphaBetaLightCurve() uses for its dynamics block. Required for
+          estimate_errors=True (see below) - Q4's L1/Nelder-Mead fit has no natural Jacobian to
+          build an analytic error estimate from.
 
     Arguments:
         v_data: [ndarray] Velocity data (m/s).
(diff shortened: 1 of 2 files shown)


wmpl/Utils/AlphaBeta.py
```diff
@@ -348,68 +398,580 @@ def _alphaBetaMinimization(x, v_normed, ht_normed):
...
+            sigma_v_init_used = sigma_v_init
+
+    else:
+        # Standard error of the median (NOT the plain MAD, which measures the dispersion of the
+        #   observations themselves, not the precision of the median as an estimator of v_init)
+        #   of the same leading points fitAlphaBeta() used to derive v_init.
+        max_index = max(int(0.2*len(v_data)), 10)
+        v_head = v_data[:max_index]
+        mad = np.median(np.abs(v_head - np.median(v_head)))
+        sigma_v_init_used = 1.2533*1.4826*mad/np.sqrt(max_index)
```
dvida: **Bug (small, real): wrong N in the sqrt.** `max_index = max(int(0.2*len(v_data)), 10)` is used as the divisor, but when `len(v_data) < 10` the slice `v_data[:max_index]` only has `len(v_data)` points, so this divides by sqrt(10) instead of sqrt(actual head size) and understates `sigma_v_init`. Measured: an 8-point trajectory gives 98.9 m/s where the correct-N value is 110.6 m/s (11% understated). For N ≥ 10 the slice length and divisor agree exactly (verified at N = 12/15/30).

```suggestion
        sigma_v_init_used = 1.2533*1.4826*mad/np.sqrt(len(v_head))
```

wmpl/Utils/AlphaBeta.py
```diff
@@ -348,68 +398,580 @@ def _alphaBetaMinimization(x, v_normed, ht_normed):
...
+            print("WARNING: v_init was given without sigma_v_init - its uncertainty will NOT be "
+                "propagated into alpha_std_rel/beta_std_rel (v_init is being treated as exactly "
+                "known).")
+            sigma_v_init_used = 0.0
+
+        else:
+            sigma_v_init_used = sigma_v_init
+
+    else:
+        # Standard error of the median (NOT the plain MAD, which measures the dispersion of the
```
dvida, line 647: **Docstring caveat needed — MC calibration shows real under-coverage on this path.** The SE-of-the-median is the right quantity for the *random* error of the derived v_init, but the median of a decelerating head is also *systematically* biased low by the deceleration trend across the window (−11 m/s on the synthetic test geometry, deterministic), and no variance term can cover a bias. Measured consequence (300-realization MC, details in the review body): with derived v_init at nominal noise the 95% gaussian CIs cover the true alpha/beta only **67/71%** of the time — while the analytic *stds* stay correct to 1-5% — recovering to ~90% at 3x noise as the fixed bias shrinks relative to the scatter. (With v_init supplied externally, coverage is a nominal 96-98%.)

The bias itself is the pre-existing v_init derivation, not this PR — but this PR's intervals are what make it consequential. Please add a caveat to the `sigma_v_init`/Statistical-caveats docstring text: with internally-derived v_init, the intervals cover only the random error and can under-cover when the deceleration across the leading-20% window is non-negligible relative to the noise; prefer an external v_init + sigma_v_init when interval coverage matters.

wmpl/Utils/AlphaBeta.py
```diff
@@ -348,68 +398,580 @@ def _alphaBetaMinimization(x, v_normed, ht_normed):
...
+    }
+
+    # Warn if the fitted (alpha, beta) landed near an optimization bound, using the same
+    #   log-space tolerance fitAlphaBetaLightCurve() uses for its own bound check (alpha/beta
+    #   span orders of magnitude, so a linear-space check would misfire - see
+    #   _makeBoundChecker()). The Gauss-Newton curvature below is only meaningful around a
+    #   genuine local optimum, and tends to be degenerate (or numerically nonsensical) once the
+    #   fit has been clipped against a box constraint instead of converging to one.
+    log_lower, log_upper = _logAlphaBetaBounds()
+    bound_tol = 1e-2
```
dvida, line 674: **Duplication with the PR-75 tip — please rebase and reuse the shared helper.** This branch predates `3969e1c` on the PR-75 branch, which extended `_makeBoundChecker()` with `log=True` for exactly this check (same tolerance, same message format — this loop is a near-verbatim duplicate of it). Verified that merging `3969e1c` into this head is conflict-free and all 20 tests pass on the merged tree; after rebasing, this whole loop reduces to:

```python
bound_warnings = []
_checkLogBound = _makeBoundChecker(bound_warnings, tol=bound_tol, log=True)
_checkLogBound("alpha", alpha, *ALPHA_BETA_BOUNDS[0])
_checkLogBound("beta", beta, *ALPHA_BETA_BOUNDS[1])
```

wmpl/Utils/AlphaBeta.py
```diff
@@ -348,68 +398,580 @@ def _alphaBetaMinimization(x, v_normed, ht_normed):
...
+
+    # Linear (alpha, beta) space, via the delta method (see the docstring above)
+    scale = np.array([alpha, beta])
+    cov = np.outer(scale, scale)*cov_log
+    alpha_std, beta_std = np.sqrt(np.diag(cov))
+    corr = cov[0, 1]/(alpha_std*beta_std)
+
+    # Parametric (log-normal) confidence interval - see the docstring above for how this differs
+    #   from wmpl.Utils.Math.confidenceInterval()'s empirical percentiles elsewhere in the library
+    z = scipy.special.ndtri(0.5 + ci/200.0)
```
dvida, line 745: **Nit: validate `ci`.** Out-of-range values degrade silently: `ci=100` → `z=inf` (CI bounds 0/inf), `ci>100` → `z=NaN`, `ci<0` → negative z (lower/upper bounds inverted). A one-line guard (`if not 0 < ci < 100: raise ValueError(...)`) at the top of `fitAlphaBeta()` alongside the `estimate_errors`/method check would do.

wmpl/Utils/AlphaBeta.py
```diff
@@ -348,68 +398,580 @@ def _alphaBetaMinimization(x, v_normed, ht_normed):
...
+            None, estimated from the MAD of the Q4 fit's residuals (floored at 1 m/s, see
+            _estimateSigmaV()) - an *effective* uncertainty that absorbs both measurement noise
+            and alpha-beta model misspecification, not a pure instrumental one.
+        loss: [str] Only used for method='robust'. scipy.optimize.least_squares loss.
+        f_scale: [float] Only used for method='robust'. Robust loss scale, in units of sigma_v.
+        estimate_errors: [bool] If True, also return a 4th element (a dict, see Return) with an
+            analytic error estimate. Only supported for method='robust' - raises ValueError
+            otherwise (see the method argument above). Off by default, so existing
+            3-tuple-unpacking call sites (v_init, alpha, beta = fitAlphaBeta(...)) are unaffected.
+        sigma_v_init: [float] Uncertainty on v_init (m/s), only used when estimate_errors=True:
```
dvida, line 816: **Nit: `sigma_v_init` passed together with `v_init=None` is silently ignored.** It's documented here, but the PR's own philosophy elsewhere is loud degradation (e.g. the warning when v_init is given without sigma_v_init). A one-line warning in the `else` branch of `_alphaBetaRobustErrors()` when `sigma_v_init is not None` would match.

dvida (changes requested):
Detailed review of the new commit (`cf791bd`) — methodology, math, numerics, and an empirical Monte Carlo calibration of the analytic errors (1500 fits total). TL;DR: the method is sound, honestly documented, and empirically well-calibrated where its assumptions hold; requesting changes for one small real bug, one docstring caveat that the MC results show is genuinely needed, and a rebase/refactor item. Inline comments mark the exact spots.

## Monte Carlo calibration (300 noise realizations per scenario, synthetic trajectory from the test suite: alpha=50, beta=2, v_init=18000 m/s, 80 pts, 85→40 km)

| Scenario | analytic/empirical std ratio (ln α, ln β) | 95% CI coverage (α / β) |
|---|---|---|
| A: v_init given exactly, noise 20 m/s | 1.02, 1.05 | 96.3% / 97.0% |
| B: v_init given exactly, noise 60 m/s | 1.02, 1.04 | 96.7% / 97.7% |
| C: v_init derived internally, noise 20 m/s | 1.01, 1.05 | **67.3% / 71.3%** |
| D: v_init derived internally, noise 60 m/s | 1.00, 1.02 | 90.0% / 90.7% |
| E: v_init perturbed N(0, 50 m/s), `sigma_v_init=50` supplied | 1.04, 1.04 | 97.3% / 98.0% |

(binomial 1-sigma on a coverage estimate ≈ 1.3%; zero NaN failures across all 1500 fits; analytic correlation matched empirical to ±0.01 everywhere)

- **A/B**: the Gauss-Newton covariance is excellently calibrated — stds within 2-5% of the empirical scatter, coverage at/marginally above nominal.
- **E**: the rank-1 `sigma_v_init` propagation does exactly its job when v_init actually fluctuates with a known, supplied sigma.
- **C/D — the one real applicability limitation**: the analytic *scatter* is still correct (ratios 1.00-1.05), but coverage collapses at low noise. The cause is **bias, not variance**: the median-of-the-leading-20% v_init estimator is biased low by the deceleration trend across the head window (−11 m/s on this geometry, deterministic), which maps to a systematic ln-beta offset of +0.033 ≈ 1.4 empirical sigmas. The auto-derived `sigma_v_init` (SE of the median, ≈ 6 m/s here) correctly covers the *random* error of the median but cannot cover a *systematic* bias — no variance term can. At 3x noise the fixed bias shrinks relative to the stds and coverage recovers to ~90%. The bias comes from the pre-existing v_init derivation, not this PR — but this PR's CIs are what make it consequential, so it needs a docstring caveat (inline comment at the auto-derivation block).

## Method verification

- Confirmed against scipy 1.17 source that `res.jac` under `loss='soft_l1'` is the robust-scaled Jacobian (`scale_for_robust_loss_function`: `J_scale = sqrt(rho' + 2 rho'' f²)`), so `s2 * inv(J^T J)` is the documented Gauss-Newton covariance of the robustified cost — the same recipe as `fitAlphaBetaLightCurve()`, keeping the module internally consistent. The docstring's "local curvature, not inverse Fisher information" caveat is accurate.
- Delta-method log→linear transform, correlation invariance, and the log-normal CI construction (`ndtri(0.5 + ci/200)`) are all correct; the `_gaussian` naming, deliberately distinct from the library's empirical-MC `confidenceInterval()`/`_95ci` convention, is a good decision.
- **FD step for the v_init sensitivity**: measured `d(ln alpha, ln beta)/d v_init` at relative steps 1e-6…1e-2 — stable to ~4 significant digits over 1e-6…1e-3, truncation error only at 1e-2. The chosen `1e-4*v_init` sits in the flat plateau; concern about optimizer-tolerance noise in a refit-based derivative is empirically unfounded.
- **Negative-diagonal guard**: reproduced the bound-pinned case and confirmed `inv(J^T J)` indeed returns without `LinAlgError` but with a negative variance — the guard is necessary and the explicit warning is a real usability fix over the old bare sqrt RuntimeWarning.
- The PSD argument for `cov_fit + cov_v0` (rank-1 term can never turn the checked diagonal negative) is correct. The ±dv refits run only after the covariance passes its guards, pass `sigma_v` through consistently, and cannot recurse.

## Tests / compatibility / style

- All 20 tests pass on the PR head, and again after merging the PR-75 tip `3969e1c` (merge is conflict-free). The 7 new tests are deterministic, match the file's style, and the bound-pinned regression test sanity-checking its own setup is well done.
- Backward compatibility is clean (default 3-tuple unchanged; `method.lower()` move is a behavioral no-op; `estimate_errors` + `'q4'` correctly rejected).
- Style matches the module (`.format()`, `Arguments:/Return:` docstrings, `_`-prefixed helper).
- Optional test gap: nothing asserts the errors are *calibrated*; a ~50-realization MC coverage smoke test would lock in the table above against regressions.

AstroEloy:
Before merging, I will add an error visualization function

wmpl/Utils/AlphaBeta.py
```diff
@@ -348,144 +409,1771 @@ def _alphaBetaMinimization(x, v_normed, ht_normed):
...
+    """
+
+    # Grid spans ALPHA_BETA_BOUNDS in LOG space (matching the space the model/optimizer/cov_log
+    #   all actually work in), via linspace + exp rather than converting to linear bounds first and
+    #   letting geomspace re-derive that same log-space grid internally - mathematically identical
+    #   (geomspace IS exp(linspace(log(...)))), just avoids a redundant log/exp round trip and
+    #   keeps the log-space nature of the computation explicit.
+    log_lo = log_bounds[0][fixed_idx] + np.log(1.001)
+    log_hi = log_bounds[1][fixed_idx] + np.log(0.999)
+    log_grid = np.linspace(log_lo, log_hi, n_grid)
```
dvida, line 1109: **High — profile CI is systematically too narrow for a well-constrained parameter at the default `n_grid`.** The grid spans the entire `ALPHA_BETA_BOUNDS` (~16 log-units for alpha) with a fixed `n_grid`, so its log spacing is ~0.065. But a well-constrained parameter's CI half-width is much smaller — for alpha on the synthetic case, `alpha_std_rel ≈ 0.0052` → CI half-width ≈ 0.010 in log, i.e. **6.3× narrower than one grid cell**. No grid point lands inside the interval, so the crossing is interpolated between the `value_hat` anchor (Δ=0) and the first grid point, where Δcost is already ≈155. Linear interpolation of that steeply-curved (≈quadratic) rise across one coarse cell badly underestimates where it crosses χ²₁.

Measured consequence (MC, 200 realizations): **alpha profile coverage 58% vs nominal 95%**; profile alpha CI width 0.47 vs Wald 1.02. Confirmed it is purely grid-limited — raising `n_grid` to 2500 recovers the width to 1.01 (≈Wald). The `value_hat` anchor makes the bracket straddle the estimate but does not fix the interpolation error.

Fix options (cheapest first): (a) build the grid *locally* around `value_hat`, scaled to the Wald std (e.g. span `value_hat*exp(±k·std_rel)`), falling back to the full bounds only when a side never crosses; (b) replace the fixed-grid search with a bracketed root-find (`scipy.optimize.brentq` on `delta(x) − chi2_1`); (c) at minimum, warn when the interval is resolved by fewer than a few grid cells and document that `n_grid` must be raised substantially for tightly-constrained parameters.

wmpl/Utils/AlphaBeta.py
```diff
@@ -616,6 +2304,10 @@ def fitAlphaBetaMass(
         density=None,
         v_init=None,
         v_final=None,
+        method='robust',
```
dvida, line 2352: **High / your call — this silently changes results for every existing `fitAlphaBetaMass` caller.** Before this PR `fitAlphaBetaMass()` had no `method` argument and always used Q4 Nelder-Mead; it now defaults to `method='robust'`, minimizing a different objective, so density/alpha/beta/mass outputs change for the same input on any call that doesn't pass `method`. It's also outside the PR's stated scope (error estimation for `fitAlphaBeta`).

The rationale is sound (the Q4 `_alphaBetaResidual` → 0 as alpha → 0 spurious minimum under `mass_constraint='final'`), and it's recoverable via `method='q4'` — but the Q4 path now has **zero test coverage** (all four mass-constraint tests call this without `method=`, so they only exercise the new default). Suggest either: (a) keep the robust default but add an explicit docstring note that results differ from the pre-PR behavior *and* add a Q4-path regression test; or (b) split this into its own PR so it's reviewed on its own merits rather than riding the error-estimation change.

wmpl/Utils/AlphaBeta.py
```diff
@@ -348,144 +409,1771 @@ def _alphaBetaMinimization(x, v_normed, ht_normed):
...
+        Falls back to re-optimizing (via _profileCostAt(), fixing beta and re-optimizing alpha
+        from its own value - the same mechanism every grid point already uses) for any other
+        loss, including a custom callable - not worth hand-rolling scipy's other built-in loss
+        formulas, or trying to support arbitrary callables generically, just to skip what is, for
+        those, already a near-free optimizer call starting exactly at its own answer.
+    """
+
+    if loss == 'soft_l1':
+        residuals = _dynResiduals(np.log([alpha, beta]), v_normed, ht_normed, sigma_v_normed)
+        z = (residuals/f_scale)**2
```
dvida, line 1080: **Medium — reference cost assumes `(alpha, beta)` is the soft_l1 minimum, but `profileAlphaBeta` advertises (and tests) working from a `method='q4'` fit.** This direct-formula branch (and the `(value_hat, 0.0)` anchor in `_profileOneParameter`) both assume the passed-in point is the joint minimum of the soft_l1 cost. A Q4 estimate minimizes a different objective, so `cost_best` is evaluated *off* the soft_l1 minimum — which inflates `s2` and makes the anchor's Δ=0 inconsistent with the true profile minimum.

Confirmed empirically: profiling beta from a Q4 fit gives min Δcost = **−0.42** (a re-optimized grid point beats `cost_best`), vs +0.23 from the robust fit; the Q4-based beta CI comes out shifted/wider ([1.949, 2.051] vs robust [1.959, 2.054]). Small on this benign synthetic case, but it biases the interval whenever the Q4 and robust optima diverge (e.g. with outliers). Suggest re-deriving `cost_best`/`value_hat` from a quick robust re-fit when `fit` isn't already a robust optimum, or warning, or tightening the docstring to require a robust point estimate with matching `loss`/`f_scale`.

wmpl/Utils/AlphaBeta.py
```diff
@@ -348,144 +409,1771 @@ def _alphaBetaMinimization(x, v_normed, ht_normed):
...
+        raise ValueError("param must be 'alpha', 'beta', or 'both', got {!r}.".format(param))
+
+    if not (0 < ci < 100):
+        raise ValueError("ci must be strictly between 0 and 100, got {!r}.".format(ci))
+
+    v_normed = v_data/v_init
+    ht_normed = ht_data/HT_NORM_CONST
+
+    if sigma_v is None:
+        sigma_v = _estimateSigmaV(alpha, beta, v_normed, ht_normed, v_init)
```
dvida, line 1378: **Low — this can break the "identical noise-scaling convention" the docstring promises.** `profileAlphaBeta` re-estimates `sigma_v` here by default, using `fit`'s own alpha/beta. If the original `fitAlphaBeta` used a different `sigma_v` (e.g. an explicit one, or one derived from a different point estimate), the two `s2` values differ and the profile interval is scaled differently from the Wald covariance it's meant to be compared against — contradicting the docstring claim (~line 1285) that "any difference is due to the cost surface's actual shape." Worth a docstring note that `sigma_v`/`loss`/`f_scale` must be threaded through identically, or thread the fit's `sigma_v` in.

wmpl/Utils/AlphaBeta.py
```diff
@@ -348,144 +409,1771 @@ def _alphaBetaMinimization(x, v_normed, ht_normed):
...
+    log_hi = log_bounds[1][fixed_idx] + np.log(0.999)
+    log_grid = np.linspace(log_lo, log_hi, n_grid)
+    grid = np.exp(log_grid)
+
+    other_bounds = (log_bounds[0][1 - fixed_idx], log_bounds[1][1 - fixed_idx])
+
+    # chi2(df=1) quantile at `ci`: exact closed form via the standard normal quantile (same
+    #   scipy.stats-avoiding trick as _gaussianEllipsePoints()/alpha_ci_wald_lower):
+    #   chi2.ppf(p, df=1) = ndtri(0.5 + p/2)^2
+    chi2_1 = scipy.special.ndtri(0.5 + ci/200.0)**2
```
dvida, line 1117: **Low — χ²(1) vs F(1, dof).** Since the deviance is divided by `s2` (an estimated scale), the self-consistent threshold is the F(1, dof) quantile; χ²(1) is its large-`dof` limit. Negligible for typical meteor point counts, but it's an approximation layered on top of the (well-documented) soft_l1↔likelihood one and isn't mentioned. One sentence in the Statistical caveats would close it.

wmpl/Utils/AlphaBeta.py
```diff
@@ -348,144 +409,1771 @@ def _alphaBetaMinimization(x, v_normed, ht_normed):
...
+    Keyword arguments:
+        errors: [dict or None] The 4th element returned by fitAlphaBeta(..., estimate_errors=
+            True), or None. If None, or if its covariance fields are NaN (e.g. the fit landed on
+            an ALPHA_BETA_BOUNDS edge - see fitAlphaBeta()'s docstring), the band/fan/ellipse
+            content degrades gracefully to an annotation instead of raising.
+        band: [bool] Draw the parameter uncertainty band in the main panel.
+        fan: [bool] Draw individual propagated-uncertainty realizations (a subsample of the same
+            draws used for the band) in the main panel.
+        band_samples: [int] Number of (alpha, beta) draws from the (ln alpha, ln beta) Gaussian
+            approximation (covariance cov_log). The forward model is cheap to evaluate, so this
```
dvida: **Low / perf — the band is not cheap, contrary to this note.** "The forward model is cheap to evaluate, so this is set by ... stability, not by cost" understates it: the band's forward-model evaluation is an un-vectorized Python loop over `band_samples` `alphaBetaVelocity()` calls, each doing a per-height scalar root-find. Timed at 80 points: band_samples=100 → 2.2 s, 1000 (default) → 8.1 s, 5000 → 36 s (linear, ~7 ms/sample) — so a default `plotAlphaBeta` call takes ~8 s. Either vectorize `alphaBetaVelocityNormed` over the sample dimension, or change this note to say the band cost scales linearly with `band_samples`.

wmpl/Utils/AlphaBeta.py
```diff
@@ -348,144 +409,1771 @@ def _alphaBetaMinimization(x, v_normed, ht_normed):
...
+        plt.show()
+
+    result_axes = tuple(ax_map[name] for name in panel_names)
+    if len(result_axes) == 1:
+        result_axes = result_axes[0]
+
+    return fig, result_axes
+
+
+def plotProfileAlphaBeta(v_data, ht_data, fit, profile, delta_cost=True, degeneracy=True,
```
dvida, line 1972: **Low / test gap — `plotProfileAlphaBeta` has no test.** It isn't imported in `test_AlphaBeta.py` and is never called, yet it has non-trivial logic (log-log interpolation of `other_profile`, open-vs-closed CI-side branching, the `_pairAt` closure with default-arg capture). Add at least a smoke/axes-contract test mirroring the `plotAlphaBeta` ones (`testPlotAlphaBetaGracefulDegradation`/`AxesContract`).

wmpl/Utils/AlphaBeta.py
```diff
@@ -348,144 +409,1771 @@ def _alphaBetaMinimization(x, v_normed, ht_normed):
...
+        (for sigma_v_init's rank-1 correction) are not free - bundling plotting into the fit
+        would force redoing that work just to redraw a figure.
+
+        Naming discipline: the shaded region in the main panel, and the words used for it
+        throughout this function, are "parameter uncertainty band"/"propagated parameter
+        uncertainty", NEVER "confidence band". It is a Monte Carlo propagation of the analytic
+        (ln alpha, ln beta) covariance through the deterministic forward model - not a classical
+        confidence band, and not the same kind of object as wmpl.Utils.Math.confidenceInterval()'s
+        empirical percentiles elsewhere in this library either, even though the same percentile
+        formula is used once the samples exist. Same discipline as the existing "_gaussian"
```
dvida: **Trivial — stale reference after the rename.** This still says `Same discipline as the existing "_gaussian" suffix on alpha_ci_wald_lower/upper`, but the fields were renamed to `_wald`. Change `"_gaussian"` → `"_wald"` (it's the only spot the rename missed; the other `_gaussian` matches are the unrelated `_gaussianEllipsePoints`).

dvida (changes requested):
Round-2 review after the updates. Thanks for the thorough work here — all five round-1 findings are cleanly resolved, and the new profile-likelihood CI is the right addition for the weakly-constrained regime.

I re-ran a Monte Carlo calibration (200 realizations/scenario) comparing the new profile CI against the Wald CI, and it turned up one real accuracy defect plus confirmed one of my code-reading concerns. Details inline; summary here.

## Round-1 findings: all resolved ✅
sqrt(len(v_head)) fix; inline bound loop now uses the shared `_makeBoundChecker(log=True)` (and you routed `fitAlphaBetaMass`/`fitAlphaBetaLightCurve` through it too — nice); `ci` validated at all three entry points; `sigma_v_init`-with-`v_init=None` now warns; and the derived-v_init under-coverage caveat is in the docstring.

## MC calibration of the profile CI (200 realizations each)

| Scenario | Param | Wald | Profile |
|---|---|---|---|
| Well-constrained (β=2, noise=20, N=80) | alpha | 96% | **58%** |
| " | beta | 94% | 93% |
| Weakly-constrained (β=2, noise=60, N=25) | alpha | 98% | 98% |
| " | beta | 74% | 100% (open side) |

- **The weakly-constrained case — the one profile CIs are for — works well.** The profile correctly reports beta as unconstrained-above (`ci_upper=None`) in all 200 realizations, the honest answer, whereas the Wald beta interval overflows `exp()` to inf (a leaked RuntimeWarning at line 749) or gives a misleading finite interval that covers only 74%.
- **But the well-constrained ALPHA profile CI is ~2x too narrow (58% coverage).** Root cause is grid resolution, confirmed below — see the inline comment on the grid construction.

## Substantive items
1. **Profile CI too narrow for well-constrained parameters at the default `n_grid`** (High) — grid-construction comment.
2. **`fitAlphaBetaMass` default silently changed q4 → robust** (High / your call) — signature comment.
3. **Profile on a `method='q4'` fit uses an off-minimum reference cost** (Medium, empirically confirmed: min Δcost = −0.42) — `_costAtOptimum` comment.

Plus minor items inline (sigma_v/s2 provenance, χ²(1) vs F(1,dof), band-plot cost vs its docstring, `plotProfileAlphaBeta` untested, a stale `_gaussian` doc reference).

Overall: merge after the two High items are addressed (or the mass-fit change split out). The statistical caveats throughout are honest and thorough — this is careful work.

dvida:
Followed up on the two substantive items with a verified patch (attached below — nothing pushed; apply with `git am`). Re-ran the MC calibration to confirm it works.

## Grid-resolution defect → fixed

Replaced the linear-interpolation crossing in `_profileOneParameter` with a bracketed root-find (`scipy.optimize.brentq` on `2(cost−cost_best)/s2 − chi2_1`, re-optimizing the profiled-out parameter inside the objective) between the `value_hat` anchor and the first grid point above threshold, with a linear-interp fallback if the bracket sign check ever fails. `n_grid` now only sets the resolution of the returned arrays, not the CI accuracy.

MC coverage (200 realizations/scenario), before → after:

| Scenario | Param | Before | After |
|---|---|---|---|
| Well-constrained (β=2, noise=20, N=80) | alpha | 58% | **96%** (Wald 96%) |
| " | beta | 93% | 94% (Wald 94%) |
| " | beta CI width vs Wald | 0.083 vs 0.124 | 0.125 vs 0.124 |
| Weakly-constrained (β=2, noise=60, N=25) | beta | open-side | open-side preserved (20/20) |

The weakly-constrained "unconstrained → `ci_upper=None`" behavior is untouched (that path returns before any interpolation).

## Other findings addressed in the same patch
- **Q4-reference cost (Medium):** warn when a profiled re-optimization beats the reference cost (`min delta_cost < 0`) — i.e. the passed-in point isn't the joint minimum of `loss` (a q4 estimate under soft_l1). Verified it fires (min Δcost −0.419) for a q4 fit, silent for a robust fit.
- **sigma_v/s2 provenance & χ²(1)-vs-F(1,dof):** documented in the `profileAlphaBeta` caveats.
- **band_samples docstring:** corrected (cost grows ~linearly, order seconds at the default 1000).
- **stale `_gaussian` doc reference:** fixed to `_wald`.
- **`fitAlphaBetaMass` default change:** documented that `method='robust'` is now the default and changes results vs the pre-PR Q4 behavior (`method='q4'` reproduces), + `testMassConstraintQ4Method` regression test.
- **`plotProfileAlphaBeta`:** added `testPlotProfileAlphaBeta` smoke/axes-contract test.

All 29 tests pass. Patch:

<details>
<summary>pr76-grid-fix.patch (git am)</summary>

```diff
From ced65cf14192b60e50f9de8942f775bc44d5072a Mon Sep 17 00:00:00 2001
From: Denis Vida <denis.vida@gmail.com>
Date: Mon, 20 Jul 2026 12:03:03 +0200
Subject: [PATCH] Fix profile-CI grid defect + address remaining review
 findings

profileAlphaBeta():
- Locate each CI edge by root-finding the profile-cost threshold crossing
  (scipy.optimize.brentq, re-optimizing the profiled-out parameter at each trial
  value) between the value_hat anchor and the first grid point above threshold,
  instead of linearly interpolating across one grid cell. The old interpolation
  was systematically too narrow whenever the true CI was narrower than the fixed
  full-bounds grid spacing - the common case for a well-constrained parameter.
  MC (200 realizations): well-constrained alpha profile-CI coverage recovers from
  58% to 96% (nominal 95%); weakly-constrained open-side (ci_upper=None) behavior
  is preserved. Falls back to linear interpolation if the bracket sign check fails.
- Warn when a profiled re-optimization beats the reference cost (min delta_cost
  < 0), i.e. the supplied (alpha, beta) is not the joint minimum of `loss` (e.g. a
  method='q4' estimate profiled under soft_l1) - the interval is then measured
  from an off-minimum reference and may be biased.
- Document that sigma_v must be passed through to match the Wald interval's s2,
  and add the F(1, dof)-vs-chi2(1) small-dof caveat.

plotAlphaBeta(): correct the band_samples docstring (cost grows ~linearly with
band_samples, order seconds at the default 1000; it is not free), and fix a stale
"_gaussian" reference left over from the _wald rename.

fitAlphaBetaMass(): document that method='robust' is now the default (a behavior
change from the pre-argument Q4-only behavior; pass method='q4' to reproduce old
results), and add testMassConstraintQ4Method() pinning the Q4 mass-fit path.

Add testPlotProfileAlphaBeta() smoke/axes-contract test (was untested).

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>
---
 wmpl/Utils/AlphaBeta.py            | 120 ++++++++++++++++++++---------
 wmpl/Utils/Tests/test_AlphaBeta.py |  68 +++++++++++++++-
 2 files changed, 151 insertions(+), 37 deletions(-)

diff --git a/wmpl/Utils/AlphaBeta.py b/wmpl/Utils/AlphaBeta.py
index 94e92d4..9db34db 100644
--- a/wmpl/Utils/AlphaBeta.py
+++ b/wmpl/Utils/AlphaBeta.py
@@ -1144,18 +1144,14 @@ def _profileOneParameter(fixed_idx, value_hat, other_hat, cost_best, s2, v_norme
 
     below, above = grid < value_hat, grid > value_hat
 
-    # Walk outward from value_hat on each side, with (value_hat, 0.0) prepended as an exact anchor
-    #   (delta_cost is 0 there by construction - cost_best IS the cost at value_hat with the other
-    #   parameter re-optimized). Anchoring on value_hat itself, rather than on whatever the fixed
-    #   geomspace grid's nearest surviving point happens to be, matters because that nearest point
-    #   can be on the WRONG side of value_hat when the grid is coarser than the true interval width
-    #   (common for a well-constrained parameter: n_grid is fixed across the full ALPHA_BETA_BOUNDS
-    #   span, so its local spacing can easily exceed a tightly-constrained parameter's actual CI) -
-    #   interpolating from there can then place ci_upper/ci_lower on the wrong side of value_hat
-    #   entirely. Walking outward also keeps delta ascending by construction on both sides, which
-    #   np.interp's docs require of `xp` and silently gives wrong answers for otherwise (the
-    #   previous ci_lower search fed it descending [delta[i], delta[i+1]], since delta strictly
-    #   decreases approaching value_hat from below).
+    # Split the grid into the two sides of value_hat, with (value_hat, 0.0) prepended as an exact
+    #   anchor (delta_cost is 0 there by construction - cost_best IS the cost at value_hat with the
+    #   other parameter re-optimized). The anchor is the inner end of the root-finding bracket
+    #   below: it is guaranteed to sit below the threshold (delta=0 < chi2_1), so it always pairs
+    #   with the first grid point above the threshold to give a valid sign-changing bracket, even
+    #   when the whole CI is narrower than one grid cell and no grid point falls inside it (the
+    #   common case for a well-constrained parameter, since n_grid is fixed across the full
+    #   ALPHA_BETA_BOUNDS span).
     log_value_hat = np.log(value_hat)
     log_grid_above = np.concatenate(([log_value_hat], log_grid[above]))
     delta_above = np.concatenate(([0.0], delta[above]))
@@ -1163,25 +1159,47 @@ def _profileOneParameter(fixed_idx, value_hat, other_hat, cost_best, s2, v_norme
     log_grid_below = np.concatenate(([log_value_hat], log_grid[below][::-1]))
     delta_below = np.concatenate(([0.0], delta[below][::-1]))
 
-    # Interpolate in LOG space, not linear grid space, then exponentiate back: the model, cov_log,
-    #   and the grid itself are all built on (ln alpha, ln beta), where the cost surface is far
-    #   closer to locally linear/parabolic than it is in linear (alpha, beta) space - interpolating
-    #   there gives a more accurate crossing point at a given grid resolution.
-    ci_upper = None
-    idx = np.where(delta_above > chi2_1)[0]
-    if len(idx):
-        i = idx[0]
-        log_ci_upper = np.interp(chi2_1, [delta_above[i - 1], delta_above[i]],
-            [log_grid_above[i - 1], log_grid_above[i]])
-        ci_upper = np.exp(log_ci_upper)
+    # Locate each CI edge by root-finding the crossing 2*(cost - cost_best)/s2 = chi2_1 in LOG
+    #   space, rather than linearly interpolating between the two bracketing grid points. Linear
+    #   interpolation of the (locally near-quadratic) delta across one grid cell systematically
+    #   MIS-estimates the crossing whenever the true CI is narrower than the grid spacing - and
+    #   because n_grid is fixed across the full ALPHA_BETA_BOUNDS span, that is the common case for
+    #   a well-constrained parameter (its CI can be many times narrower than one cell), where the
+    #   only bracket available is [value_hat (delta=0), first grid point (delta>>chi2_1)] and the
+    #   linear estimate comes out roughly 2x too narrow. A bracketed root-find between the
+    #   value_hat anchor (delta=0 < chi2_1 by construction) and the first grid point above the
+    #   threshold re-optimizes the profiled-out parameter at each trial value, so the reported edge
+    #   is accurate to the solver tolerance regardless of grid resolution (n_grid then only sets
+    #   the resolution of the returned 'grid'/'delta_cost' arrays and the initial bracketing scan).
+    def _delta_minus_threshold(log_value):
+        cost, _ = _profileCostAt(np.exp(log_value), fixed_idx, np.log(other_hat), v_normed,
+            ht_normed, sigma_v_normed, other_bounds, loss, f_scale)
+        return 2.0*(cost - cost_best)/s2 - chi2_1
+
+    def _crossing(log_side, delta_side):
+        # First grid point (index i, i >= 1 since delta_side[0] is the anchor's 0.0) whose delta
+        #   exceeds the threshold; None if the grid never crosses (the data does not constrain
+        #   this side within ALPHA_BETA_BOUNDS). The root is bracketed by the value_hat anchor
+        #   (delta 0, so _delta_minus_threshold < 0) and that grid point.
+        idx = np.where(delta_side > chi2_1)[0]
+        if not len(idx):
+            return None
 
-    ci_lower = None
-    idx = np.where(delta_below > chi2_1)[0]
-    if len(idx):
         i = idx[0]
-        log_ci_lower = np.interp(chi2_1, [delta_below[i - 1], delta_below[i]],
-            [log_grid_below[i - 1], log_grid_below[i]])
-        ci_lower = np.exp(log_ci_lower)
+        try:
+            log_ci = scipy.optimize.brentq(_delta_minus_threshold, log_value_hat, log_side[i],
+                xtol=1e-8)
+        except ValueError:
+            # Re-optimizing inside _delta_minus_threshold warm-starts from other_hat rather than
+            #   from the swept trajectory, so it can land marginally below the threshold right at
+            #   log_side[i] and break the bracket sign. Fall back to interpolating the (ascending)
+            #   swept deltas - never worse than the original linear estimate.
+            log_ci = np.interp(chi2_1, [delta_side[i - 1], delta_side[i]],
+                [log_side[i - 1], log_side[i]])
+        return np.exp(log_ci)
+
+    ci_upper = _crossing(log_grid_above, delta_above)
+    ci_lower = _crossing(log_grid_below, delta_below)
 
     return {
         'ci_lower': ci_lower,
@@ -1251,7 +1269,11 @@ def profileAlphaBeta(v_data, ht_data, fit, sigma_v=None, ci=95.0, param='both',
         sigma_v: [float] Velocity uncertainty (m/s), same role as fitAlphaBeta()'s sigma_v. If
             None, estimated the same way (_estimateSigmaV(), using `fit`'s own alpha/beta as the
             reference model) - the same *effective*, not purely instrumental, uncertainty
-            fitAlphaBeta() itself would derive.
+            fitAlphaBeta() itself would derive. NOTE: to compare this interval against
+            fitAlphaBeta(estimate_errors=True)'s Wald interval on the identical noise scaling (s2),
+            pass the SAME sigma_v (and loss/f_scale) here that produced that fit - left to default,
+            the re-estimated sigma_v can differ from the fit's, giving a different s2 and hence a
+            differently-scaled interval.
         ci: [float] Confidence level (0-100, e.g. 95, the default).
         param: [str] 'alpha', 'beta' (most relevant given beta is the parameter this matters for
             in this model - see above), or 'both' (default).
@@ -1312,7 +1334,12 @@ def profileAlphaBeta(v_data, ht_data, fit, sigma_v=None, ci=95.0, param='both',
           the same well-motivated but not perfectly theoretically clean comparison this module
           already accepts elsewhere for its Gauss-Newton covariances (e.g.
           fitAlphaBetaLightCurve()'s alpha_std_rel/beta_std_rel) - a useful, standard engineering
-          approximation, not a literal likelihood-ratio test.
+          approximation, not a literal likelihood-ratio test. There is a second, milder
+          approximation on top of that: because the delta is scaled by the estimated s2 (below),
+          the self-consistent threshold is the F(1, dof) quantile rather than chi2(df=1); the two
+          coincide only in the large-dof limit (chi2(df=1) = dof*F(1, dof) as dof -> inf), so for
+          the point counts typical here (dof in the tens to hundreds) the difference is small, but
+          the interval is very slightly optimistic at small dof.
         - Not a replacement for estimate_errors=True's cov_log/corr_log/alpha_std_rel/
           beta_std_rel - those remain the right tool for LOCAL error propagation (e.g. via the
           delta method into some other derived quantity). This answers a different question
@@ -1371,6 +1398,21 @@ def profileAlphaBeta(v_data, ht_data, fit, sigma_v=None, ci=95.0, param='both',
         profile['beta'] = _profileOneParameter(1, beta, alpha, cost_best, s2, v_normed,
             ht_normed, sigma_v_normed, log_bounds, ci, n_grid, loss, f_scale)
 
+    # cost_best is the cost at the passed-in (alpha, beta), used as the profile's reference
+    #   minimum. If a profiled-out re-optimization along the grid ever reaches a LOWER cost
+    #   (delta_cost < 0), then (alpha, beta) was not actually the joint minimum of this loss -
+    #   which happens when `fit` came from a different objective than `loss` (e.g. a method='q4'
+    #   point estimate profiled under the default soft_l1): the reference is then off-minimum, so
+    #   the interval is measured from an inflated baseline and can be biased/mis-centered. Warn
+    #   rather than silently reporting it (pass a robust `fit` with matching loss/f_scale to avoid).
+    min_delta = min(float(np.min(p['delta_cost'])) for p in profile.values())
+    if min_delta < -1e-6:
+        print("WARNING: a profiled re-optimization reached a lower cost than the reference "
+            "(min delta_cost = {:.3g} < 0), so the supplied (alpha, beta) is not the joint "
+            "minimum of loss='{:s}'. The profile CI is measured from an off-minimum reference and "
+            "may be biased - pass a fit from fitAlphaBeta(method='robust') with the same "
+            "loss/f_scale.".format(min_delta, loss))
+
     if verbose:
         print()
         print("--- profileAlphaBeta ---")
@@ -1588,7 +1630,7 @@ def plotAlphaBeta(v_data, ht_data, fit, errors=None, band=True, fan=True, band_s
         (ln alpha, ln beta) covariance through the deterministic forward model - not a classical
         confidence band, and not the same kind of object as wmpl.Utils.Math.confidenceInterval()'s
         empirical percentiles elsewhere in this library either, even though the same percentile
-        formula is used once the samples exist. Same discipline as the existing "_gaussian"
+        formula is used once the samples exist. Same discipline as the existing "_wald"
         suffix on alpha_ci_wald_lower/upper.
 
     Arguments:
@@ -1605,8 +1647,11 @@ def plotAlphaBeta(v_data, ht_data, fit, errors=None, band=True, fan=True, band_s
         fan: [bool] Draw individual propagated-uncertainty realizations (a subsample of the same
             draws used for the band) in the main panel.
         band_samples: [int] Number of (alpha, beta) draws from the (ln alpha, ln beta) Gaussian
-            approximation (covariance cov_log). The forward model is cheap to evaluate, so this
-            is set by how stable the tail percentiles of the band need to be, not by cost.
+            approximation (covariance cov_log). Set this by how stable the tail percentiles of the
+            band need to be, but note the cost is NOT negligible: each draw evaluates the forward
+            model (a per-height numerical velocity inversion) once, un-vectorized, so runtime grows
+            roughly linearly with band_samples (order a few seconds at the default 1000 for a
+            typical trajectory - reduce it for quick interactive plots).
         fan_curves: [int] Number of individual curves drawn for `fan`, subsampled from the same
             `band_samples` draws (capped at however many landed inside ALPHA_BETA_BOUNDS).
         ellipse: [bool] Draw the (ln alpha, ln beta) ellipse panel.
@@ -2353,7 +2398,12 @@ def fitAlphaBetaMass(
             model (or, under mass_constraint="initial", of the final fitted model).
         method: [str] 'robust' (default) or 'q4' - fitting approach used both for the preliminary
             unconstrained alpha-beta fit (see Notes below) and for the (density, beta[, mu])
-            optimization under the selected mass_constraint.
+            optimization under the selected mass_constraint. NOTE: 'robust' is the default, which
+            is a change from this function's original behavior - before the fitting `method`
+            argument existed, this function always used the 'q4' path. Existing callers that do
+            not pass `method` will therefore get (generally slightly) different fitted
+            density/alpha/beta/mass values than before. Pass method='q4' to reproduce the
+            pre-change results exactly.
               - 'q4': the classic Gritsevich (2007) Q4 fit throughout (minimizeAlphaBeta() for the
                 preliminary fit; scipy.optimize.minimize(method='Nelder-Mead') on
                 _alphaBetaResidual()'s transformed, height-weighted residual for the constrained
diff --git a/wmpl/Utils/Tests/test_AlphaBeta.py b/wmpl/Utils/Tests/test_AlphaBeta.py
index c9f57f8..c8f88f6 100644
--- a/wmpl/Utils/Tests/test_AlphaBeta.py
+++ b/wmpl/Utils/Tests/test_AlphaBeta.py
@@ -25,8 +25,8 @@ import matplotlib.pyplot as plt
 from wmpl.Utils.AlphaBeta import (fitAlphaBetaMass, fitAlphaBeta, fitAlphaBetaLightCurve,
     alphaBetaMasses, alphaBetaVelocity, alphaBetaHeight, alphaBetaVelocityNormed,
     alphaBetaHeightNormed, alphaBetaLuminosityF, alphaBetaModelMagnitude,
-    alphaBetaLuminousEfficiency, plotAlphaBeta, profileAlphaBeta, _profiledMagOffset,
-    _gaussianEllipsePoints, HT_NORM_CONST, P_0M, ALPHA_BETA_BOUNDS)
+    alphaBetaLuminousEfficiency, plotAlphaBeta, plotProfileAlphaBeta, profileAlphaBeta,
+    _profiledMagOffset, _gaussianEllipsePoints, HT_NORM_CONST, P_0M, ALPHA_BETA_BOUNDS)
 
 
 # True parameters used to generate the synthetic trajectory. The height range is chosen so the
@@ -197,6 +197,34 @@ def testBothMassConstraint():
         assert 0 <= mu_best <= 2/3
 
 
+def testMassConstraintQ4Method():
+    """ The legacy Q4 fitting path of fitAlphaBetaMass() (method='q4', the behavior before the
+        `method` argument was added, when robust became the default) must still recover the true
+        parameters. Regression guard for the now-non-default Q4 path, which the other mass tests
+        no longer exercise since they use the robust default.
+    """
+
+    v_data, v_final_true, m_init_true, m_final_mu0_true, m_final_mu23_true = _syntheticTrajectory()
+
+    for density in [DENS_TRUE, None]:
+
+        res = fitAlphaBetaMass(v_data, HT_DATA, SLOPE, m_init_true, mass_constraint="initial", \
+            density=density, v_init=V_INIT_TRUE, method='q4', verbose=False)
+
+        (v_init, v_final,
+            density_mu0, alpha_mu0, beta_mu0, m_initial_mu0, m_final_mu0,
+            density_mu23, alpha_mu23, beta_mu23, m_initial_mu23, m_final_mu23,
+            mu_best, density_mu_best, alpha_mu_best, beta_mu_best, m_initial_mu_best,
+            m_final_mu_best) = res
+
+        _assertClose(alpha_mu0, ALPHA_TRUE, 0.05, "alpha_mu0 (q4)")
+        _assertClose(beta_mu0, BETA_TRUE, 0.05, "beta_mu0 (q4)")
+        _assertClose(density_mu0, DENS_TRUE, 0.05, "density_mu0 (q4)")
+
+        # The imposed initial mass must still be reproduced exactly
+        _assertClose(m_initial_mu0, m_init_true, 1e-6, "m_initial_mu0 (q4)")
+
+
 def testDerivedVelocities():
     """ When v_init and v_final are not given, they should be derived from the data. """
 
@@ -775,6 +803,40 @@ def testProfileAlphaBetaRejectsBadInputs():
         raise AssertionError("profileAlphaBeta: ValueError not raised for param='banana'")
 
 
+def testPlotProfileAlphaBeta():
+    """ plotProfileAlphaBeta() must draw a full multi-panel figure without raising, honor the
+        single-panel `axes` contract (reuse the passed-in axes), and reject `axes` for a
+        multi-panel request - mirroring plotAlphaBeta()'s axes contract.
+    """
+
+    v_data, _, _, _, _ = _syntheticTrajectory()
+
+    fit = fitAlphaBeta(v_data, HT_DATA, v_init=V_INIT_TRUE, method='robust')
+    profile = profileAlphaBeta(v_data, HT_DATA, fit)
+
+    # Full multi-panel figure must build and return (fig, ax_map)
+    fig, ax_map = plotProfileAlphaBeta(v_data, HT_DATA, fit, profile)
+    plt.close(fig)
+
+    # A single-panel request (one parameter, one panel type) with axes must reuse them
+    profile_alpha = profileAlphaBeta(v_data, HT_DATA, fit, param='alpha')
+    fig_in, ax_in = plt.subplots()
+    fig_out, ax_out = plotProfileAlphaBeta(v_data, HT_DATA, fit, profile_alpha,
+        degeneracy=False, trajectories=False, axes=ax_in)
+    assert fig_out is fig_in and ax_out is ax_in
+    plt.close(fig_in)
+
+    # Multi-panel + axes must raise
+    try:
+        plotProfileAlphaBeta(v_data, HT_DATA, fit, profile, axes=ax_in)
+
+    except ValueError:
+        pass
+
+    else:
+        raise AssertionError("plotProfileAlphaBeta: ValueError not raised for multi-panel + axes")
+
+
 def testLuminosityModel():
     """ The luminosity function f(v) must be positive on the interior, vanish at both ends, and
         the model magnitude must respond to the offset additively.
@@ -938,6 +1000,7 @@ if __name__ == "__main__":
         testInitialMassConstraint,
         testFinalMassConstraint,
         testBothMassConstraint,
+        testMassConstraintQ4Method,
         testDerivedVelocities,
         testInputValidation,
         testAlphaBetaNormedRoundTrip,
@@ -956,6 +1019,7 @@ if __name__ == "__main__":
         testPlotAlphaBetaRejectsBadCi,
         testProfileAlphaBetaBracketsEstimate,
         testProfileAlphaBetaRejectsBadInputs,
+        testPlotProfileAlphaBeta,
         testLuminosityModel,
         testProfiledMagOffsetWeighted,
         testFitAlphaBetaLightCurve,
-- 
2.43.0

```

</details>

dvida:
@AstroEloy heads up — I pushed a `master` sync to this branch (merge commit `246a101`), now that #75 and #77 are in `master`. It's a plain merge commit, conflict-free (both `master` and this branch had independently added the log-space option to `_makeBoundChecker()`, and git reconciled the identical edits with no duplication), and all tests pass on the merged tree.

**Please `git pull` before your next push** so your local branch doesn't diverge from the updated remote.

The grid-fix patch I posted above is separate and still yours to take or adapt — it applies cleanly on top of this sync.

dvida (approved):
Looks good — approving.

- The grid-resolution fix is in (verified: well-constrained coverage recovers 58% → 96%, weakly-constrained open-side preserved), along with the Q4-reference warning, the `fitAlphaBetaMass` default note + Q4 regression test, the sigma_v/χ² caveats, the band-plot docstring, and the `plotProfileAlphaBeta` test.
- Removing the non-rigorous `alpha_std_rel`/`beta_std_rel`/`mu_std` from `fitAlphaBetaLightCurve` is the right call — it was the least defensible of the estimates, and `fitAlphaBeta(estimate_errors=True)` + `profileAlphaBeta()` now cover errors properly. Verified no callers/tests depend on the removed keys.

All 29 tests pass; branch is synced to master and merges clean. One note for the release notes: this drops those three return-dict fields from `fitAlphaBetaLightCurve` (shipped in #75), so it's a small breaking change to that function's output. The remaining error work (rigorous light-curve/mass uncertainties, band-plot vectorization) is fine as follow-up.

Issue #71, opened 2026-07-14
Labels containing mathtex throw a warning for unrecognised escape sequences


markmac99:
In python 3.11 or later, plain strings containing mathtex symbols throw a SyntaxWarning for unrecognised escape sequences. 

There are a few of these in Analysis/ and TrajSim/ - i will work on a PR to correct it. 

``` python
sign=1
ref_point=10
ref_point_std=0.56
label='Reference point = {:.2f} $\pm$ {:.2f}'.format(sign*ref_point, ref_point_std)
<stdin>:1: SyntaxWarning: invalid escape sequence '\p'
print(label)
Reference point = 10.00 $\pm$ 0.56
```

The fix is to make the string an r-string (raw string), as this stops python trying to interpret `\` as introducing an escape sequence. 
``` python
label=r'Reference point = {:.2f} $\pm$ {:.2f}'.format(sign*ref_point, ref_point_std)
print(label)
Reference point = 10.00 $\pm$ 0.56
```

markmac99:
looks like you can also escape the backslash to achieve the same result. This is being done elsewhere so i will use the same approach. 
``` python
label='Reference point = {:.2f} $\\pm$ {:.2f}'.format(sign*ref_point, ref_point_std)
print(label)
Reference point = 10.00 $\pm$ 0.56
```

dvida:
PR #73 is merged — thanks Mark! It fixes all the warnings in the three files it touches.

For completeness, I compiled the whole `wmpl` package with warnings enabled and there are 9 remaining invalid escape sequence warnings in other files, in case you'd like to address them in a follow-up:

- `wmpl/TrajSim/ShowerSim.py:846` — `\%`
- `wmpl/MetSim/MetSim.py:1345` — `\D`
- `wmpl/MetSim/FitSimAnalyzer.py:346` — `\i`
- `wmpl/Trajectory/CorrelateRMS.py:974, 981, 988` — `\d` (these are regex patterns, so they should become raw strings `r"..."` rather than double-backslash)
- `wmpl/Dynesty/DynestyMetSim.py:2425, 2428, 2455, 2460` — `\d`, `\%`, `\_`

I'll leave this issue open until those are cleaned up.

markmac99:
I'm on the case :)

markmac99:
Closed via PR #77

PR #77, opened 2026-07-16
Fixing remaining mathtex and regexp syntax errors

markmac99:
Fixed issues as mentioned in issue #71 . 

Note that i am unable to test the simulations as i do not have any suitable data . I've run the builtin test cases, where present, however `DynestyMetSim` is still running after ~4 hours so i don't know if its working yet!

### DIFF
--- a/wmpl/Dynesty/DynestyMetSim.py
+++ b/wmpl/Dynesty/DynestyMetSim.py
@@ -1761,7 +1761,7 @@ def plotDynestyResults(dynesty_run_results, obs_data, flags_dict, fixed_values, 
         'disruption_mass_max_ratio': r"$m_{max}/m_{disr}$",
         'disruption_mass_grain_ratio': r"$m_{gr}/m_{disr}$",
         'height': r"$h$ [km]",
-        'mass_percent': r"$m_{percent}$ [\%]",
+        'mass_percent': r"$m_{percent}$ [%]",
         'number': r"$N$",
         'sigma': r"$\sigma$ [kg/MJ]",
         'erosion_coeff': r"$\eta$ [kg/MJ]",
@@ -1794,7 +1794,7 @@ def plotDynestyResults(dynesty_run_results, obs_data, flags_dict, fixed_values, 
         'disruption_mass_max_ratio': r"$m_{max}/m_{disr}$",
         'disruption_mass_grain_ratio': r"$m_{gr}/m_{disr}$",
         'height': r"$h$ [m]",
-        'mass_percent': r"$m_{percent}$ [\%]",
+        'mass_percent': r"$m_{percent}$ [%]",
         'number': r"$N$",
         'sigma': r"$\sigma$ [kg/J]",
         'erosion_coeff': r"$\eta$ [kg/J]",
@@ -2422,7 +2422,7 @@ def plotDynestyResults(dynesty_run_results, obs_data, flags_dict, fixed_values, 
 
     """
         for i, label in enumerate(labels):
-            coverage_val = "\ding{51}" if coverage_mask[i] else "\ding{55}"  # Use checkmark/x for coverage
+            coverage_val = r"\ding{51}" if coverage_mask[i] else r"\ding{55}"  # Use checkmark/x for coverage
             latex_str += (f"    {label} & {lower_95[i]:.4g} & {truths[i]:.4g} & {best_guess_table[i]:.4g} & {approx_modes[i]:.4g} "
                         f"& {posterior_mean[i]:.4g} & {posterior_median[i]:.4g} & {upper_95[i]:.4g} "
                         f"& {abs_error[i]:.4g} & {rel_error[i]:.4g}\% & {coverage_val} \\\\\n") #\hline\n
@@ -2452,12 +2452,12 @@ def plotDynestyResults(dynesty_run_results, obs_data, flags_dict, fixed_values, 
 
     # check if the file_name has a _ in it if so put \ before it
     if '_' in file_name:
-        file_name_caption = file_name.replace('_', '\_')
+        file_name_caption = file_name.replace('_', r'\_')
     else:
         file_name_caption = file_name
 
     if hasattr(obs_data, 'const'):
-        latex_str += f"Posterior summary statistics for {file_name_caption} test case. The Best Fit is the simulation with the highest likelihood. Absolute and relative errors are calculated based on the Best Fit. The Cover column indicates whether the true value lies within the 95\% CI."
+        latex_str += f"Posterior summary statistics for {file_name_caption} test case. The Best Fit is the simulation with the highest likelihood. Absolute and relative errors are calculated based on the Best Fit. The Cover column indicates whether the true value lies within the 95% CI."
     else:
         latex_str += f"Posterior summary statistics for {file_name_caption} meteor. The Best Fit is the simulation with the highest likelihood."
     latex_str += r"""}
--- a/wmpl/Dynesty/DynestyMetSim.py
+++ b/wmpl/Dynesty/DynestyMetSim.py
@@ -1761,7 +1761,7 @@ def plotDynestyResults(dynesty_run_results, obs_data, flags_dict, fixed_values, 
         'disruption_mass_max_ratio': r"$m_{max}/m_{disr}$",
         'disruption_mass_grain_ratio': r"$m_{gr}/m_{disr}$",
         'height': r"$h$ [km]",
-        'mass_percent': r"$m_{percent}$ [%]",
+        'mass_percent': r"$m_{percent}$ [\%]", 
         'number': r"$N$",
         'sigma': r"$\sigma$ [kg/MJ]",
         'erosion_coeff': r"$\eta$ [kg/MJ]",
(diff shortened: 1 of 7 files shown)


wmpl/Dynesty/DynestyMetSim.py
```diff
@@ -1761,7 +1761,7 @@ def plotDynestyResults(dynesty_run_results, obs_data, flags_dict, fixed_values,
         'disruption_mass_max_ratio': r"$m_{max}/m_{disr}$",
         'disruption_mass_grain_ratio': r"$m_{gr}/m_{disr}$",
         'height': r"$h$ [km]",
-        'mass_percent': r"$m_{percent}$ [\%]",
+        'mass_percent': r"$m_{percent}$ [%]",
```
dvida: This one needs to stay escaped. `variable_map` is dual-use: besides plot labels, its entries become rows of the LaTeX table written to the `.tex` file (the `for i, label in enumerate(labels)` loop around line 2424, via `labels = [variable_map[variable] ...]`). In real LaTeX a bare `%` starts a comment, so a fragmentation `mass_percent` row would have its remaining columns and the `\\` row terminator commented out, breaking the table.

Also note both maps were already raw strings, so neither of these two lines ever produced a warning — this hunk changes rendering, not warnings. The matching change in `variable_map_plot` (line 1797) is fine to keep, since that map is only used for matplotlib labels where the backslash showed up literally.

```suggestion
        'mass_percent': r"$m_{percent}$ [\%]",
```

wmpl/Dynesty/DynestyMetSim.py
```diff
@@ -2452,12 +2452,12 @@ def log_transf(v):
...
     # check if the file_name has a _ in it if so put \ before it
     if '_' in file_name:
-        file_name_caption = file_name.replace('_', '\_')
+        file_name_caption = file_name.replace('_', r'\_')
     else:
         file_name_caption = file_name
 
     if hasattr(obs_data, 'const'):
-        latex_str += f"Posterior summary statistics for {file_name_caption} test case. The Best Fit is the simulation with the highest likelihood. Absolute and relative errors are calculated based on the Best Fit. The Cover column indicates whether the true value lies within the 95\% CI."
+        latex_str += f"Posterior summary statistics for {file_name_caption} test case. The Best Fit is the simulation with the highest likelihood. Absolute and relative errors are calculated based on the Best Fit. The Cover column indicates whether the true value lies within the 95% CI."
```
dvida: `latex_str` is written verbatim to a `.tex` file a few lines below and compiled by LaTeX, so the `%` must stay escaped in the *output* — unescaped, it comments out the rest of the line including the closing `}` of `\caption{` (appended on the same line just below), and the table no longer compiles.

The original code actually produced correct output (`\%` passes through a non-raw f-string unchanged); only the Python warning was the problem. Doubling the backslash fixes the warning while keeping the output identical:

```suggestion
        latex_str += f"Posterior summary statistics for {file_name_caption} test case. The Best Fit is the simulation with the highest likelihood. Absolute and relative errors are calculated based on the Best Fit. The Cover column indicates whether the true value lies within the 95\\% CI."
```

wmpl/Dynesty/DynestyMetSim.py
```diff
@@ -2422,7 +2422,7 @@ def log_transf(v):
 
     """
         for i, label in enumerate(labels):
-            coverage_val = "\ding{51}" if coverage_mask[i] else "\ding{55}"  # Use checkmark/x for coverage
+            coverage_val = r"\ding{51}" if coverage_mask[i] else r"\ding{55}"  # Use checkmark/x for coverage
             latex_str += (f"    {label} & {lower_95[i]:.4g} & {truths[i]:.4g} & {best_guess_table[i]:.4g} & {approx_modes[i]:.4g} "
                         f"& {posterior_mean[i]:.4g} & {posterior_median[i]:.4g} & {upper_95[i]:.4g} "
                         f"& {abs_error[i]:.4g} & {rel_error[i]:.4g}\% & {coverage_val} \\\\\n") #\hline\n
```
dvida: This is the one invalid-escape warning left in the repo after this PR (`invalid escape sequence '\%'`). Same situation as the caption: the LaTeX output genuinely needs `\%`, so the fix is to double the backslash rather than remove it. (A raw f-string won't work here because the trailing `\n` needs to stay a real newline.)

```suggestion
                        f"& {abs_error[i]:.4g} & {rel_error[i]:.4g}\\% & {coverage_val} \\\\\n") #\hline\n
```

wmpl/Trajectory/CorrelateRMS.py
```diff
@@ -971,21 +971,21 @@ def yearMonthDayDirInDtRange(self, dir_name):
             date_fmt = "%Y"
 
             # Check if the directory name starts with a year
-            if not re.match("^\d{4}", dir_name):   # noqa: W605 
+            if not re.match(r"^\d{4}", dir_name):   # noqa: W605
```
dvida: Non-blocking nit: now that these three patterns are raw strings, the `# noqa: W605` comments here and on the two matches below are stale and can be dropped.

dvida (changes requested):
Thanks for finishing off #71 — I verified this by compiling every `.py` file on both `master` and this branch and recording invalid-escape warnings. Master currently has 11; this branch gets it down to 1 (see inline comment at `DynestyMetSim.py:2428`).

The regex raw-strings, the `time.clock()` -> `time.process_time()` replacement, the `$V_{\\infty}$` fix, and the `ShowerSim` repr change are all correct. The `MetSim.py` scatter-label rework slightly changes the rendering (the `=` and units move outside math mode) but that's cosmetic and fine.

However, two of the `%` un-escapes in `DynestyMetSim.py` land in a string that is written to a `.tex` file and compiled by real LaTeX (not matplotlib mathtext), where a bare `%` starts a comment and breaks the table — see inline comments. Requesting changes for those two plus the one remaining warning.

No worries about not being able to test `DynestyMetSim` end-to-end — the affected code only runs in the results/plotting stage, and the inline comments below don't require running a simulation to verify.

wmpl/Trajectory/CorrelateRMS.py
markmac99: > Non-blocking nit: now that these three patterns are raw strings, the `# noqa: W605` comments here and on the two matches below are stale and can be dropped.

good point, i will remove them

wmpl/Dynesty/DynestyMetSim.py
markmac99: understood - will fix it correctly.

wmpl/Dynesty/DynestyMetSim.py
markmac99: understood.  Been a while since i wrote any LaTeX. :)

wmpl/Dynesty/DynestyMetSim.py
markmac99: got it. I have LaTeX installed somewhere so if i can get the test case to run successfully i will check it out.

markmac99:
By the way i also had to make two changes in TrajConversions.py at lines 869 and 872 to make sure we were passing arrays of floats to rotateVector, which passes the arrays to vectMag, which has been numbified and Numba threw an error that it can only process arrays of floats or complex. Of course [0,1,0] is an array of ints. 
Similarly at line 57 of MetSim.py i had to convert state_vect from a tuple to an array to avoid the same error in vectNorm. 

If i remove the numba decorations from vectNorm and vectMag, the problem goes away but i guess we want to keep the benefits of numba here.

wmpl/TrajSim/TrajSim.py
```diff
@@ -54,7 +54,7 @@ def _radiantDiff(radiant_eq, ra_g, dec_g, v_init, state_vector, jd_ref):
 
 
     # Assume that the velocity at infinity corresponds to the initial velocity
-    v_init = np.sqrt(v_g**2 + (2*6.67408*5.9722)*1e13/vectMag(state_vector))
+    v_init = np.sqrt(v_g**2 + (2*6.67408*5.9722)*1e13/vectMag(np.array(state_vector)))
```
dvida: Non-blocking: `np.array()` keeps the input's dtype, so an integer `state_vector` would still hit the numba `TypingError` (numba's `np.linalg.norm` only supports float arrays — I confirmed `vectMag(np.array([1, 2, 3]))` fails). Forcing the dtype makes it robust to any input:

```suggestion
    v_init = np.sqrt(v_g**2 + (2*6.67408*5.9722)*1e13/vectMag(np.array(state_vector, dtype=np.float64)))
```

dvida (approved):
All three requested changes are correctly applied — I re-ran the compile scan over every `.py` file on the updated branch and there are now **zero** invalid-escape warnings in the repo. I also verified the generated LaTeX strings render correctly (`\%` in the table row and caption, `\ding{51}` intact).

The two numba-related fixes check out too: `vectMag` is `@njit`-compiled and numba's `np.linalg.norm` raises a `TypingError` for both plain lists and integer arrays (confirmed locally), so the float literals in `rotatePolar` and the `np.array()` wrap in `TrajSim.py` are both necessary. One optional hardening suggestion inline.

Thanks for the quick turnaround!

wmpl/TrajSim/TrajSim.py
dvida: Applied in d9eeeff — verified `vectMag` now works with int lists, int arrays, and float arrays.

markmac99: Unless you already did it i will propose the dtype tweak via a separate update, thanks for the suggestion.

PR #78, opened 2026-07-21
Add LUT-accelerated Alpha-Beta velocity inversion and fast pipeline

AstroEloy:
- Added an experimental LUT-based implementation of alphaBetaVelocityNormed() using a PCHIP inverse-Ei lookup table.
- Added a fast=True option and propagated it through the full Alpha-Beta fitting pipeline (fitAlphaBeta, fitAlphaBetaMass, fitAlphaBetaLightCurve, profileAlphaBeta, and plotting utilities).
- Exposed getDefaultInverseEiLUT() to allow one-time LUT pre-warming and reuse.
- Added validation tests covering numerical agreement with the original brentq implementation, API propagation, and pipeline-level speedups.
- Benchmarks show ~500× acceleration for the standalone inversion, >3× for fitAlphaBeta(method="robust"), and >10× for profileAlphaBeta(), while preserving fit results within the tested tolerances.
- Documented the existing numerical precision limitation of the forward formulation for β ≳ 30 due to floating-point cancellation (unchanged in this PR).

### DIFF
--- a/wmpl/Utils/AlphaBeta.py
+++ b/wmpl/Utils/AlphaBeta.py
@@ -14,6 +14,7 @@ import sys
 import numpy as np
 import scipy.special
 import scipy.optimize
+import scipy.interpolate
 
 
 from wmpl.Utils.Math import meanAngle
@@ -441,7 +442,7 @@ def _logAlphaBetaBounds():
     )
 
 
-def _estimateSigmaV(alpha0, beta0, v_normed, ht_normed, v_init):
+def _estimateSigmaV(alpha0, beta0, v_normed, ht_normed, v_init, fast=False):
     """ Robust (MAD-based) velocity uncertainty (m/s) from the residuals of a Q4 fit
         (alpha0, beta0), floored at 1 m/s. Used by minimizeAlphaBetaRobust()/fitAlphaBeta() and
         fitAlphaBetaLightCurve() whenever sigma_v isn't given explicitly.
@@ -457,21 +458,26 @@ def _estimateSigmaV(alpha0, beta0, v_normed, ht_normed, v_init):
         ht_normed: [ndarray] Height normalized to HT_NORM_CONST.
         v_init: [float] Initial velocity (m/s), used to convert the residuals to physical units.
 
+    Keyword arguments:
+        fast: [bool] Passed through to alphaBetaVelocityNormed() - see its docstring.
+
     Return:
         sigma_v: [float] Effective velocity uncertainty (m/s), floored at 1 m/s.
     """
 
-    v_model0 = alphaBetaVelocityNormed(ht_normed, alpha0, beta0)
+    v_model0 = alphaBetaVelocityNormed(ht_normed, alpha0, beta0, fast=fast)
     res0 = (v_model0 - v_normed)*v_init
     sigma_v = 1.4826*np.median(np.abs(res0 - np.median(res0)))
 
     return max(sigma_v, 1.0)
 
 
-def _dynResiduals(x, v_normed, ht_normed, sigma_v_normed):
+def _dynResiduals(x, v_normed, ht_normed, sigma_v_normed, fast=False):
     """ Dynamics residual block: model velocity residuals (v_model - v_obs), sigma-normalized.
         Used both as the dynamics half of _jointResiduals() (fitAlphaBetaLightCurve()) and, on
-        its own, by minimizeAlphaBetaRobust().
+        its own, by minimizeAlphaBetaRobust(). This is the innermost per-iteration hot path of
+        every 'robust' fit (called once per scipy.optimize.least_squares iteration), so fast=True
+        here is where alphaBetaVelocityNormedLUT() actually pays off.
 
     Arguments:
         x: [ndarray] Fit parameters (ln alpha, ln beta).
@@ -479,18 +485,21 @@ def _dynResiduals(x, v_normed, ht_normed, sigma_v_normed):
         ht_normed: [ndarray] Height normalized to HT_NORM_CONST.
         sigma_v_normed: [float] Normalized velocity uncertainty.
 
+    Keyword arguments:
+        fast: [bool] Passed through to alphaBetaVelocityNormed() - see its docstring.
+
     Return:
         res: [ndarray] Sigma-normalized velocity residuals.
     """
 
     alpha, beta = np.exp(x)
-    v_model = alphaBetaVelocityNormed(ht_normed, alpha, beta)
+    v_model = alphaBetaVelocityNormed(ht_normed, alpha, beta, fast=fast)
 
     return (v_model - v_normed)/sigma_v_normed
 
 
 def minimizeAlphaBetaRobust(v_normed, ht_normed, sigma_v_normed, loss='soft_l1', f_scale=2.0,
-        x0_alpha_beta=None, return_fit_result=False):
+        x0_alpha_beta=None, return_fit_result=False, fast=False):
     """ Robust alternative to minimizeAlphaBeta(): least squares (soft-L1 by default) directly on
         the velocity residuals (v_model - v_obs), instead of Q4's transformed residual.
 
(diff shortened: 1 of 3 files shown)


dvida (commented):
Reviewed the diff and the existing brentq path. The numerical approach is correct, well-documented, and fully backward-compatible (`fast` defaults to `False` everywhere; Q4 never inverts velocity so it's untouched). Nice work.

**Verified by hand / spot-checks**
- Derivation is sound — confirmed numerically that `Ei(β) − 2α·exp(β−y) == Ei(βv²)`, so `v = sqrt(Ei⁻¹(Ei(β) − 2α·exp(β−y))/β)` inverts Eq. (7) exactly.
- Clip semantics match the brentq path (`y ≥ y_hi → v_hi`, `y ≤ y_lo → v_lo`; `y` is monotone increasing in `v`).
- The trailing `fast` in each `least_squares(args=(...))` tuple lines up with each residual's new trailing `fast=False` param (`_dynResiduals`, `_lightCurveResiduals`, `_jointResiduals`, free-mu variants).
- Overflow of `delta = 2α·exp(β−y)` only happens inside the `y ≤ y_lo` clipped region — wrapped in `errstate(over='ignore')` and overwritten by the clip.
- The `β ≥ 10` asymptotic branch (`x = β − δβe^{−β}`) is a good catch — it handles a real near-`v=1` cancellation the pure table would otherwise suffer.

**Please address before merge**

1. **Stale/contradictory docstring.** `alphaBetaVelocityNormedLUT()`'s docstring says *"Not wired into any fitting path; kept alongside `alphaBetaVelocityNormed()` for benchmarking/validation before it replaces it there."* — but this PR does wire it into every fitting path via `fast=True`. Please update it so it doesn't contradict the feature it ships.

2. **Timing-based `assert`s will be flaky in CI.** `testAlphaBetaVelocityNormedLUT` (`speedup > 20`) and `testFastPipelineSpeedup` (`> 3`, `> 10`) are wall-clock ratios; on a shared/loaded runner they can dip below even generous thresholds and turn the build red for no real regression. The correctness tests (`< 1e-5` agreement, `testFastFlagPropagation`) are the real guarantees. Suggest keeping the speed checks but making them non-fatal — `warnings.warn`/skip instead of `assert`, or gate behind an env flag / `pytest.mark`.

**Optional / minor**

3. `fast` is threaded through ~20 signatures — wider API surface, but consistent with how the module already threads `loss`/`f_scale`/`sigma_v`, so fine.
4. `fast=True` degrades silently (clips the table domain) for β ≳ 30. Documented in the module comment and shared with the brentq path, so acceptable — a one-line note in the public `fast` docstrings that results degrade rather than error there would be a small plus.
5. Interior `v` can marginally exceed `1 − v_eps` when interpolation gives `x` slightly `> β` away from the `y ≥ y_hi` clip — bounded by the tested `< 1e-5` error, so negligible; noting only.

Overall: approve once items 1–2 are addressed.

wmpl/Utils/AlphaBeta.py
```diff
@@ -1563,6 +1611,170 @@ def alphaBetaVelocity(ht_data, alpha, beta, v_init, v_eps=1e-10):
...
+    return _default_inverse_ei_lut
+
+
+def alphaBetaVelocityNormedLUT(ht_normed, alpha, beta, v_eps=1e-10, lut=None):
+    """ EXPERIMENTAL - LUT-accelerated alternative to alphaBetaVelocityNormed(). Inverts Eq. (7)
+        in closed form using a tabulated inverse of Ei() instead of bracketing
+        alphaBetaHeightNormed() with brentq per point - see the module comment above for the
+        derivation and known limitations. Not wired into any fitting path; kept alongside
+        alphaBetaVelocityNormed() for benchmarking/validation (test_AlphaBeta.py
+        ::testAlphaBetaVelocityNormedLUT) before it replaces it there.
```
dvida: The docstring contradicts the PR: `fast=True` does wire this into every fitting path. Suggested rewording:

```suggestion
        derivation and known limitations. Used by every fast=True code path in this module
        (alphaBetaVelocityNormed(..., fast=True) dispatches here); kept as a separate function,
        alongside the brentq-based alphaBetaVelocityNormed(), for direct benchmarking/validation
        (test_AlphaBeta.py::testAlphaBetaVelocityNormedLUT).
```

dvida (commented):
Attempted the two fix items as suggested changes so you can accept or tweak them inline. Suggestion 1 (docstring) is on `AlphaBeta.py`; suggestions 2-4 turn the flaky timing `assert`s into non-fatal warnings on the test file (using `print` so no new import is needed). Feel free to reword.

wmpl/Utils/Tests/test_AlphaBeta.py
```diff
@@ -345,6 +348,188 @@ def testAlphaBetaNormedRoundTrip():
...
+    t0 = time.perf_counter()
+    alphaBetaVelocityNormedLUT(ht_normed, ALPHA_TRUE, BETA_TRUE)
+    t_lut = time.perf_counter() - t0
+
+    t0 = time.perf_counter()
+    alphaBetaVelocityNormed(ht_normed, ALPHA_TRUE, BETA_TRUE)
+    t_brentq = time.perf_counter() - t0
+
+    speedup = t_brentq/t_lut
+    assert speedup > 20, "LUT speedup {:.1f}x <= 20x".format(speedup)
```
dvida: Wall-clock ratio -> flaky on loaded CI. Downgrade to a non-fatal warning (accuracy asserts above stay the real guarantee):

```suggestion
    # Soft check: timing ratios are machine/CI-load dependent, so a miss warns rather than fails
    #   the suite (the <1e-5 accuracy asserts above are the real guarantees).
    if speedup <= 20:
        print("WARNING: LUT speedup {:.1f}x <= 20x (perf only, not a correctness failure)".format(
            speedup))
```

wmpl/Utils/Tests/test_AlphaBeta.py
```diff
@@ -345,6 +348,188 @@ def testAlphaBetaNormedRoundTrip():
...
+    t_slow = time.perf_counter() - t0
+
+    t0 = time.perf_counter()
+    fit = fitAlphaBeta(v_data, HT_DATA, v_init=V_INIT_TRUE, method='robust', verbose=False, \
+        fast=True)
+    t_fast = time.perf_counter() - t0
+
+    speedup = t_slow/t_fast
+
+    assert speedup > 3, "fitAlphaBeta fast=True speedup {:.1f}x <= 3x".format(speedup)
```
dvida: Same flaky-timing concern -> non-fatal warning:

```suggestion
    if speedup <= 3:
        print("WARNING: fitAlphaBeta fast=True speedup {:.1f}x <= 3x (perf only)".format(speedup))
```

wmpl/Utils/Tests/test_AlphaBeta.py
```diff
@@ -345,6 +348,188 @@ def testAlphaBetaNormedRoundTrip():
...
+    t0 = time.perf_counter()
+    profileAlphaBeta(v_data, HT_DATA, fit, param='beta', n_grid=40, verbose=False, fast=False)
+    t_slow = time.perf_counter() - t0
+
+    t0 = time.perf_counter()
+    profileAlphaBeta(v_data, HT_DATA, fit, param='beta', n_grid=40, verbose=False, fast=True)
+    t_fast = time.perf_counter() - t0
+
+    speedup = t_slow/t_fast
+    assert speedup > 10, "profileAlphaBeta fast=True speedup {:.1f}x <= 10x".format(speedup)
```
dvida: Same flaky-timing concern -> non-fatal warning:

```suggestion
    if speedup <= 10:
        print("WARNING: profileAlphaBeta fast=True speedup {:.1f}x <= 10x (perf only)".format(
            speedup))
```

dvida:
Heads-up @AstroEloy: I pushed one extra commit to this branch (78999ce) — it excludes the `Tests` packages from the `import wmpl` submodule walk in `wmpl/__init__.py`.

Unrelated to the LUT feature itself, but I hit it while testing this branch: because `import wmpl` imports every submodule, it was executing `test_AlphaBeta.py`'s top-level `matplotlib.use("Agg")`, which switched the whole process to the non-interactive Agg backend and broke `plt.show()` for `python -m wmpl.Utils.AlphaBeta`. The bug is on `master` too, so it can also be split into its own PR if you'd rather keep this one focused — your call. If you force-push, just be aware this commit would be dropped.

AstroEloy:
Thanks @dvida. Looking good!

PR #80, opened 2026-07-21
REBOUND: reuse Earth epoch state across MC runs (speed up MC)

dvida:
## Problem
Monte Carlo runs were slow because the Earth's epoch position was resampled for **every** realization. `convertToBarycentric` recomputed it on each call, and with the `--horizons` source that is **one web query per realization** (mc=100 → ~100 queries).

## Fix
The Earth's barycentric state at the epoch is identical for the nominal solution and all MC realizations. Compute it **once** — read directly from the Earth already added to the simulation (before `move_to_com`, so it is still barycentric; no extra query for either ephemeris source) — and pass it into `convertToBarycentric` via a new `earth_state` argument.

- Horizons source: per-run Earth queries drop from `1 + mc_runs` to **1**.
- Local source: eliminates the redundant per-realization recompute.

## Verification
- Nominal orbital elements unchanged (bit-identical to pre-change local run).
- `convertToBarycentric(..., earth_state=ref)` is bit-identical (0.0 diff) to recomputing internally.
- A forced 25-realization MC run (local) completes in ~5 s with realizations varying as expected (final `a` spread 10.91–11.07); nominal `a` unchanged at 10.984663.

### DIFF
--- a/wmpl/Rebound/REBOUND.py
+++ b/wmpl/Rebound/REBOUND.py
@@ -274,7 +274,8 @@ def detectCloseEncounters(sim_outputs, n_hill=3.0):
     return encounters
 
 
-def convertToBarycentric(state_vect, jd, log_file_path="", ephem_source="local", jpl_ephem_data=None):
+def convertToBarycentric(state_vect, jd, log_file_path="", ephem_source="local", jpl_ephem_data=None,
+                         earth_state=None):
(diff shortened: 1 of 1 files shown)


dvida:
Superseded by #81, which includes this Earth-epoch-state reuse plus the multicore Monte Carlo work. Closing in favor of #81.

PR #83, opened 2026-07-21
Fix silent state-overwrite bugs in MetSimErosion.py

AstroEloy:
Found and fixed several bugs in the reference RK4 erosion/disruption engine (MetSimErosion.py) where per-tick state was silently overwritten or excluded, producing wrong values without raising any error. Each fix was verified against simulation runs. This is an ongoing cleanup pass — expect more commits.

Bugs fixed
- erosion_height_change's density/ablation-coefficient change was re-applied every tick instead of once, silently undoing any later manual sigma change (e.g. a complex-fragmentation "A" event) on the very next tick. Added a one-time guard flag on the fragment.
- mass_total_active summed frag.m directly instead of frag.m*n_grains, undercounting total active mass as soon as any fragment represented more than one grain — inconsistent with how lum_total/electron_density_total were already weighted.
- lum_eroded/tau_eroded stayed at 0.0 whenever fragmentation_show_individual_lcs was False (the default), even though that same light was already flowing into lum_total. The aggregate tracking was accidentally nested inside the flag meant only to gate the expensive per-entry breakdown.
- brightest_height/brightest_vel/leading_frag_* excluded a fragment's own death tick from candidacy, so a fragment's last valid position/brightness was silently dropped — most visibly, a single non-eroding body reported brightest_height=0.0 on its final row instead of its real height. Also fixed a related scale bug where the stored "brightest" value used the unweighted per-grain luminosity instead of the n_grains-weighted total, letting far too many spurious hand-offs occur between candidates.
- disruption_erosion_coeff (assigned to a disruption daughter at birth) was unconditionally overwritten by the generic height-based erosion coefficient on the very next tick, making the parameter almost a no-op. Added a flag to exclude disruption daughters from that recompute.
- The "ablate what's left of the whole mass" clamp (when computed mass loss would exceed the fragment's remaining mass) used the wrong formula and did not actually floor the mass at zero — for large overshoots it could even drive mass negative, which would propagate as NaN into that tick's luminosity (log of a negative mass). Confirmed triggered ~24 times in a normal test scenario (always in near-death grains), now correctly floors at exactly 0.

Documented but intentionally not changed (design trade-offs, flagged for a future decision)
- Fixed-step RK4 (dt=0.005s) is under-resolved for grains near erosion_mass_min (can overshoot true velocity by tens of percent in a single step). Documented in ablateAll()'s docstring.
- Rate terms (mass loss, deceleration) are evaluated with the previous tick's state, while luminous/ionization efficiency use the already-updated current-tick state — a small (~1.6%) but systematic inconsistency. Documented inline where it happens.

Other findings, annotated (not fixed — nothing to fix)
- energy_per_cs_before_erosion / energy_per_mass_before_erosion are declared as output fields but never populated anywhere. The computation that would populate them exists (energyReceivedBeforeErosion()) but is never called. Annotated in place for whoever picks this up next.
- Fixed a few unrelated typos ("daguhter" -> "daughter", "occured" -> "occurred").

Testing/tooling added
- wmpl/MetSim/MetSimErosion_old.py: a frozen pre-fix snapshot of this file (git HEAD), kept only as a diffing target — not maintained going forward.
- wmpl/MetSim/Tests/test_MetSimErosion.py: a manual visual-inspection tool that runs a complex scenario (two-phase erosion + EF/A/D fragmentation + disruption) through both the fixed and pre-fix engines and saves a 3x3 comparison plot, used to verify each fix's effect end-to-end.

Extra: Excluded Tests subpackages (e.g. Utils.Tests, MetSim.Tests) from wmpl/__init__.py's automatic submodule walk, so `import wmpl` can no longer trigger import-time side effects (like matplotlib.use("Agg")) from test modules.


dvida (commented):
## Review — correctness ✅, validated on real data + synthetic scenario

Went through all six fixes against the surrounding code and the RK4 integrator, then ran both engines (`MetSimErosion_old.py` vs the fixed `MetSimErosion.py`) on 13 fitted 2019 Orionids and on the synthetic disruption/complex scenario. **All six fixes are real bugs and the fixes are correct.** The visible light curve, dynamics, and wake are unchanged on real fits — differences are confined to the faint tail and to the (now-corrected) bookkeeping arrays.

### Per-fix confirmation

| # | Fix | Verification |
|---|-----|-------------|
| 1 | `erosion_height_change` rho/sigma applied once | Synthetic scenario: main-fragment luminosity now diverges upward after the "A" event instead of being re-clobbered each tick. ✔ |
| 2/4 | brightest/leading: include death tick + `n_grains`-weighted `frag.lum` | Pre-fix stored per-grain `lum` while comparing weighted `frag.lum`, and ran *after* the `killFragment`+`continue`. On all 13 Orionids the pre-fix `brightest_height≈0` death-tick row (count **1 → 0**) disappears; synthetic run shows the pre-fix brightest-velocity oscillation storm vanish. ✔ |
| 3 | `lum_eroded`/`tau_eroded` no longer gated by `fragmentation_show_individual_lcs` | Default (`False`): peak `lum_eroded` **0.0 → 187–610 W** across all 13, matching light already in `lum_total`. ✔ |
| 5a | `mass_total_active` weighted by `n_grains` | Now consistent with `lum`/`q`; synthetic total-active-mass sits higher through mid-flight as expected. ✔ |
| 5b | `disruption_child` guard | Disruption daughters (`complex=False`) reached the `getErosionCoeff()` reassignment and lost their `disruption_erosion_coeff`; guard fixes it. `spawn_child` copies the flag to grains, but grains have `erosion_enabled=False` so never hit the reassignment → harmless. ✔ |
| 6 | mass-clamp floors at 0 | Pre-fix `mass_loss_total += frag.m` gives `m_new = 2·frag.m + loss` — e.g. `m=1, loss=-3 → m_new=-1`, still negative → `log(mass)` NaN. Reachable since `massLossRK4` alone can return `-(4/3)·m`. `= -frag.m` floors exactly at 0. ✔ |

RK4 math itself is unchanged and correct: `massLossRK4 = (k1+2k2+2k3+k4)/6`, `decelerationRK4` returns mean `dv/dt`, and `lum = -τ(½(dm/dt)v² + m·v·dv/dt) = -τ·dEk/dt`. The documented ~1.6% operator-split inconsistency (rate terms on pre-step state, efficiencies on post-step state) is a modeling choice — agree with leaving it, just documented.

### Validation on 13 fitted Orionids

- Time-aligned **total-luminosity Δmag < 0.02 on 11/13**; the two outliers (7.9, 1.9 mag) are entirely in the **faint tail** (mag > 8, negligible flux) — peak magnitude and the visible curve are identical.
- **Wake profiles identical** old vs new.
- Dynamics: the fix removes the pre-fix death-tick velocity artifact; curves track the observed points.

### Housekeeping — pushed to this branch (48e2c12)

I went ahead and pushed one follow-up commit for the cleanup we'd otherwise just be asking for:

1. **Removed `MetSimErosion_old.py`** — the 1678-line `# TO BE REMOVED` duplicate is gone from the branch.
2. **Removed the committed `Tests/complex_scenario_old_vs_new.png`** and gitignored `*.png` under `MetSim/Tests`.
3. **Rewrote `Tests/test_MetSimErosion.py` into a real pytest module** — same complex scenario (2-phase erosion + EF/A/D + disruption), new engine only, with asserts that lock in the fixes: no negative mass, no NaN/inf outputs, `lum_eroded` tracked with the default flag, `brightest_height` never a spurious 0 while luminous, and `mass_total_active >= main_mass`. So there's now a regression guard instead of a manual visual diff against a shipped copy of the old code.

Please sanity-check that commit — happy to adjust the scenario/asserts to taste.

One note to carry forward: `mass_total_active` (#5a) is a behavioral change to an output field, so any downstream consumer of it should be aware.

### Follow-up (separate PR)

We'll take the optimizations above (dead-fragment pruning is the big one) in a separate change with its own before/after validation — out of scope here.

### Optional optimizations (out of scope for this correctness PR — separate follow-up)

Verified against the current hot loop; none block this PR:

- **Dead fragments are never pruned from `fragments`.** `killFragment()` only sets `active=False`, and `fragments += frag_children_all` grows the list monotonically, so every tick re-scans every fragment ever created — the main loop, the `mass_total_active` sum, and the wake pass are each `O(total_fragments)` per tick. For long/complex runs this dominates. A periodic compaction (or a maintained active-list) is the biggest available win.
- **Fold the `mass_total_active` scan into the main loop** (accumulate after the kill-check) instead of the separate full pass at the end — removes one `O(N)` scan per tick.
- **`frag_entry = next((x for x in const.fragmentation_entries if x.id == frag.complex_id), None)`** is a linear scan per complex grain per tick; a `{id: entry}` dict built once makes it `O(1)`. Only matters when `fragmentation_show_individual_lcs=True`.
- Minor: `heightCurvature()` uses `np.sqrt` on scalars in the per-fragment path — `math.sqrt` avoids the numpy-scalar overhead.

Nice catches on the silent overwrites — these are exactly the kind of bugs that never raise but quietly bias results.

dvida:
### Performance pass folded in (8f87e0c)

Pushed the `ablateAll` hot-loop optimizations we'd discussed as a follow-up. **Behavior-preserving** — the per-tick output arrays are bit-for-bit identical to the previous commit across all 13 fitted Orionids and the synthetic erosion/fragmentation/disruption scenario (`rtol=1e-9`), so no fit results change; only the amount of work per tick does.

- **Prune fully-ablated grains from the fragment list each tick.** The list previously grew monotonically (`killFragment()` only flips `active=False`), so every tick re-scanned every grain ever created — the main loop, the `mass_total_active` sum, and the wake pass were each `O(total_fragments)`. Dead *non*-grains (the main fragment, complex-fragmentation daughters) are kept, since `runSimulation()` still scans them after the run to find the main fragment and aggregate per-entry final masses. This is the dominant win.
- **Accumulate `mass_total_active` inside the main loop** instead of a separate post-loop scan of the whole list.
- **Index complex fragmentation entries by id** (dict) once per tick, instead of a linear scan of all entries per complex grain per tick.
- **`math.sqrt`/`math.cos` on the scalar `heightCurvature` path** instead of numpy, to avoid numpy-scalar overhead in the per-fragment inner loop.

Net: **~1.96× faster** on the grain-heavy synthetic scenario (761 ms vs 1489 ms, best of 5). The regression test added in the previous commit passes on the optimized engine.

The identity check (optimized vs pre-opt engine over all 14 scenarios) and the per-meteor before/after plots are available if useful — happy to attach.

PR #84, opened 2026-07-21
Analytic error propagation and meteorite-survival diagram to AlphaBeta

AstroEloy:
- Extend alphaBetaMasses() with estimate_errors=True, mirroring fitAlphaBeta()'s own API. Given the (ln alpha, ln beta) covariance from fitAlphaBeta(estimate_errors=True), analytically propagates it (delta method) into m_init/m_final, along with optional independent uncertainties on slope, density, and (for the full alpha-beta solution) vel_init/vel_end. Returns relative/absolute std, a 2x2 (ln m_init, ln m_final) covariance and correlation, and log-normal Wald confidence intervals. Degrades to NaN cleanly when the input covariance is NaN (e.g. a bound-pinned fit), instead of raising.

- Add the _alphaBetaMassErrors() helper implementing the propagation: alpha/density/slope enter ln(m_init) and ln(m_final) exactly log-linearly, so their contribution is propagated without additional approximation; beta (and vel_init/vel_end, full solution only) enter through a genuinely nonlinear term and are linearized locally, the same caveat as fitAlphaBeta()'s own beta_ci_wald_lower/upper.

- Add plotAlphaBetaSurvivalDiagram(): the classic Gritsevich ln(alpha*sin(slope)) vs ln(beta) meteorite-survival diagram. Boundary curves for arbitrary mass thresholds are derived live from alphaBetaMasses()'s own physics at the mu=0/mu=2/3 shape-spin extremes, instead of hardcoded constants. Shades likely/possible/unlikely-fall zones for a primary threshold, draws additional thresholds as reference lines, and optionally overlays a Gaussian uncertainty ellipse built from fitAlphaBeta()'s cov_log.

- Refactor the module's legacy CLI script ("PLOT METEORITE DROPPING POSSIBILITY") to call plotAlphaBetaSurvivalDiagram() instead of its own inline hardcoded boundary constants, which did not track the --dens/--ga CLI arguments and silently drew inconsistent boundaries whenever those were overridden.

- Add 12 tests covering the new error propagation (validation, NaN propagation, point-estimate invariance, covariance structure, an exact mass-ratio identity, and a 20k-draw Monte Carlo cross-check) and the new plot (boundary-curve math, input validation, graceful degradation, axes reuse). Full suite: 44/44 passing.

### DIFF
--- a/wmpl/Utils/AlphaBeta.py
+++ b/wmpl/Utils/AlphaBeta.py
@@ -2589,7 +2589,10 @@ def _alphaBetaMassErrors(alpha, beta, slope, mu, dens, m_init, m_final, vel_init
     m_final_std_rel = np.sqrt(var_m_final)
 
     if (m_init_std_rel > 0) and (m_final_std_rel > 0):
-        corr_log = cov_m/(m_init_std_rel*m_final_std_rel)
+        # Clip to [-1, 1]: for a valid (PSD) cov_log this is already in range, but a
+        #   caller-supplied cov_log that isn't actually PSD could push it slightly outside, which
+        #   would be a nonsensical correlation coefficient - clamp rather than propagate it.
+        corr_log = np.clip(cov_m/(m_init_std_rel*m_final_std_rel), -1.0, 1.0)
     else:
         corr_log = np.nan
 
@@ -3986,13 +3989,38 @@ def plotAlphaBetaSurvivalDiagram(
         ax.text(0.02, 0.02, "Uncertainty ellipse unavailable\n(cov_log is NaN)", \
             transform=ax.transAxes, ha='left', va='bottom', fontsize=8, color='0.4')
 
-    ax.scatter(
-        [x_point],
-        [y_point],
-        c='blue',
-        zorder=5,
-        label=point_label,
-    )
+    if have_cov:
+
+        # Marginal 1-sigma error bars on the point (sqrt of cov_log's diagonal): a quick-read
+        #   complement to the ellipse above, which additionally shows the (alpha, beta)
+        #   correlation the bars alone cannot. slope is treated as exact, so sigma_x = sigma_ln
+        #   alpha and sigma_y = sigma_ln beta directly.
+        sigma_x = np.sqrt(errors['cov_log'][0, 0])
+        sigma_y = np.sqrt(errors['cov_log'][1, 1])
+
+        ax.errorbar(
+            [x_point],
+            [y_point],
+            xerr=[sigma_x],
+            yerr=[sigma_y],
+            fmt='o',
+            color='blue',
+            ecolor='tab:blue',
+            elinewidth=1.0,
+            capsize=3,
+            capthick=1.0,
+            zorder=5,
+            label=point_label,
+        )
+
+    else:
+        ax.scatter(
+            [x_point],
+            [y_point],
+            c='blue',
+            zorder=5,
+            label=point_label,
+        )
 
     ax.set_xlim(x_range)
     ax.set_ylim(y_range)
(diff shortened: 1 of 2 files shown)


dvida (commented):
**Rigorous review — LGTM, no correctness blockers.**

Verified the delta-method math end-to-end against the point-estimate formulas:
- `ln(m_init) = C − 2·ln(dens) − 3·ln(α) − 3·ln(sinθ)` and all Jacobian coefficients (`−3`, `−3·cotθ`, `−2/dens`, `−kβ`, and the two velocity coefficients for the full solution) are correct.
- `J·cov_log·Jᵀ` assembly is right, including the shared slope/dens term entering both variances **and** the cross-covariance.
- The `Var(ln(m_final/m_init)) = (kβ·β_std_rel)² + vel_terms` identity holds algebraically (α/slope/dens cancel out of the difference).
- The uncertainty ellipse uses `cov(ln α, ln β)` correctly, since `x = ln α + ln sinθ` is a constant shift.

Dependencies all present in base (`scipy.special.ndtri`, `fitAlphaBeta`'s `cov_log` convention, `_gaussianEllipsePoints` reuse). Tests are thorough — the 20k-draw Monte-Carlo cross-check is a strong correctness gate.

Two optional (non-blocking) hardening nits:
1. `corr_log` isn't clamped to [−1, 1]; a non-PSD caller-supplied `cov_log` with a positive diagonal could exceed it (the fit's own `cov_log` is PSD, so low risk). Consider a clip, or a docstring note that PSD input is assumed.
2. The Monte-Carlo test samples slope/dens from unbounded Gaussians; a tail draw with slope ≤ 0 / > π/2 or dens ≤ 0 would make `alphaBetaMasses` return NaN or raise. Negligible at σ_slope = 1° over 20k draws, but a `np.clip` on the samples would remove any flakiness risk.

Nice work — self-consistent boundary curves and graceful NaN degradation throughout.

dvida:
Pushed a follow-up commit (`1751a89`) with the two review nits fixed plus the CLI/plot wiring so the new error estimation is actually reachable from `python -m wmpl.Utils.AlphaBeta`, which it wasn't before (the `__main__` block never set `estimate_errors=True`).

**Nits fixed:**
- Clamp `corr_log` to [−1, 1] in `_alphaBetaMassErrors()` (guards a non-PSD caller-supplied `cov_log`).
- Clip the Monte-Carlo test's slope/dens samples to `alphaBetaMasses()`'s valid domain (removes any tail-draw flakiness).

**CLI wiring (`__main__`):**
- `--errors` / `-e`: forces the robust fit (the only one that produces a covariance), propagates the fitted `(ln α, ln β)` covariance into `m_init`/`m_final`, and prints 1σ relative errors + Wald CIs + `corr(α, β)`. Default (no flag) keeps the fast `q4` path and identical output.
- `--slopeunc` (deg) and `--densunc` (kg/m³): optional independent uncertainties folded into the mass error, off by default.
- Velocities deliberately left exact — `vel_end` is derived from the fitted α/β, so its uncertainty is already carried by `cov_log`; an independent `sigma_vel_*` would double-count it.

**Plots (only with a finite `cov_log`, degrade cleanly otherwise):**
- Survival diagram: marginal 1σ error bars on the point, alongside the existing correlation ellipse.
- α-β velocity-fit panel: α/β annotated with ± in the legend, plus a sampled 1σ velocity band.

Verified end-to-end on a real 9-station event and the full suite is 44/44. Note `--errors` switches the fit method `q4 → robust` (unavoidable — `q4` has no Jacobian), so α/β shift slightly vs the default; the reported point estimate is then the robust one.

AstroEloy:
Thanks!

PR #92, opened 2026-07-27
REBOUND: exact encounters, impact detection, chaos diagnostics, radiation forces, reproducible MC

dvida:
Seven commits on `wmpl/Rebound/REBOUND.py`, plus a new test module. Every quantitative claim below was measured, not assumed.

## 1. Encounter distances are exact, not sampling-limited

Encounters were found by scanning the **sampled output**, which is far too coarse near the Earth and Moon. Measured on a real trajectory:

| | |
|---|---|
| Sample spacing (500 outputs / 60 d) | 2.89 h |
| Object's motion relative to the Moon per sample | up to **167,000 km** |
| Moon's entire 3 R_Hill detection sphere | **184,000 km** |

The object can cross nearly the whole lunar sphere between two samples. Verified error of the old approach: **464 km** at `n_outputs=500`, **59,524 km** at `n_outputs=100`.

Closest approaches are now recorded by a **REBOUND heartbeat callback at every internal integrator timestep**. Cost is nil (~1,600 steps over 60 d; 0.65 s vs 0.73 s baseline), and IAS15 shrinks its timestep *during* an encounter, so resolution concentrates exactly where it matters. `detectCloseEncounters` is kept but marked deprecated.

## 2. Impact detection

`collision = "line"` + `collision_resolve = "halt"` detects a real hit along the line of travel between timesteps. Previously `collision = "none"` with inert radii — an object could pass straight through the Earth or Moon unnoticed.

- Armed only **after** the object leaves the Earth's neighbourhood (it starts at the surface).
- Radii corrected to physical values (Earth 6371 km; **Moon 1737.4 km — was 6.9% too small**), overridable via `body_radii` for an atmospheric-entry cross-section.
- Verified: a particle aimed at Earth records `impact: Earth at t = 0.4868 d` and truncates the run.

## 3. The Moon must be gated differently (geometry, verified exactly)

A lunar encounter requires being within **0.00394 AU** of Earth, while the Earth-departure gate is **0.02943 AU** — 7.5x larger. So every lunar encounter happens *inside* the gate, on the outbound leg or a return. The Moon is therefore tracked over the **whole** integration while all other bodies are gated to post-departure; gating the Moon identically would make lunar encounters structurally undetectable.

Useful consequence: Earth detection arms exactly at 3 R_Hill, so an Earth encounter can only be flagged on a **genuine return**.

## 4. Chaos diagnostics that don't lie

The divergence estimate reuses the MC ensemble (neighbouring trajectories) — no extra variational particles, free when MC is already running.

Three real defects were found and fixed by self-review:
- **One truncated realization destroyed the whole estimate.** `min()` across all realizations meant a single impact-truncated run collapsed the fit. Reproduced: one 3-sample realization out of 20 gave `None`. Each realization now contributes over its own length.
- **Overconfident classification.** A 3000-day run reported *"Lyapunov time = 980.79 days"* from **R2(exp) = 0.603 vs R2(lin) = 0.507** — two equally poor fits — while detecting **no close encounters at all**. Now requires R2 >= 0.90 plus a 0.05 margin.
- **Rejecting long runs outright** and demanding a shorter re-run (wasting a 400 s integration). It now fits the **compact leading window** already present in the data.

Then the deeper problem: **position separation is the wrong metric.** On a 36,000-day run the cloud saturated after ~1200 d while the only encounter was at ~21,000 d — saturation was ordinary **Keplerian phase drift** (tiny delta-a gives different periods, so the cloud smears along the track), which occurs whether or not the orbit is chaotic. So **semi-major-axis divergence** was added, which is immune to phase drift:
- 60-day run: sigma_a factor **0.99**, correctly reported *regular*, with the x465 position spread explained as phase drift.
- 36,000-day run: sigma_a factor **2.03**, growth neither exponential nor linear, so weakly chaotic and **no timescale claimed**.

The exponential branch (previously never executed by any run) recovers synthetic e-folding times of 20, 50 and 150 days exactly.

## 5. Radiation pressure and Poynting-Robertson (opt-in)

Via `--beta`, or `--radius`/`--density` which compute `beta = 5.7425e-4*Q_pr/(rho*s)`. Verified: 1 micron grain at 3000 kg/m^3 gives beta = 0.19 (matches literature), 1 cm meteoroid 1.9e-5, and a 10 micron grain shifts the 60-day semi-major axis from 1.4549 to 1.4031 AU. Off by default, since a trajectory solution does not constrain size or density.

**Yarkovsky is deliberately excluded.** REBOUNDx exposes it, but it runs silently with unset thermal/spin parameters rather than erroring, and a fireball's spin axis, rotation period and thermal conductivity are unknowable for a body that no longer exists — it would yield precise-looking numbers resting on invented inputs.

## 6. Bug fixes and reporting

- **`--forward N` silently ignored N**, integrating `--days` (60) instead, despite its help text; bare `--forward` errored. Now takes an optional day count.
- **MC was not reproducible** — no RNG seeding, so identical invocations gave different error bars. Added `--seed`; when absent a seed is drawn and **reported** so any run can be reproduced.
- **Plot crashed** (`x and y must have same first dimension`) whenever an MC realization differed in length from the nominal run. Each series now uses its own time axis.
- **`HILL_RADII_AU` was internally inconsistent** by up to 5.7% versus its own masses. Regenerated from one documented convention, `r_H = a*(m/3M_sun)^(1/3)`.
- **Energy drift** diagnostic (2.45e-10, healthy for IAS15). An initial 1.2e-3 came from comparing energies across two different frames — caught and fixed.
- **Escape detection** via `exit_max_distance`, so a runaway object is flagged instead of integrated forever.
- **Tisserand parameter** and dynamical class. Validated: 2P/Encke 3.0265 (lit. ~3.03), 1P/Halley -0.6066 (~-0.6), Ceres 3.3103.
- The integration span is now stated **before** the run and reported as the span *actually achieved*, flagging a shortfall if an impact truncated it.
- `--outputs` to control sample density; outer-planet distance axis starts at zero; `pkg_resources` and empty-legend warnings silenced; peri/node label swap fixed (`peri` = REBOUND `omega`, `node` = `Omega`, matching `Orbit.peri`/`Orbit.node`).

## 7. Tests

New `wmpl/Rebound/Tests/test_REBOUND_numerics.py` — **23 tests**, all passing, covering encounter thresholds and ordering, departure gating, divergence (including the truncation and saturation cases), element divergence, Tisserand, radiation beta, and Hill-table consistency. They are pure numerics, so they run without REBOUND/REBOUNDx installed. The 3 pre-existing dependency tests still pass.

## Compatibility

`reboundSimulate()` keeps its two-value return by default; diagnostics only with `return_diagnostics=True`. Reference case nominal elements unchanged: a = 1.454858, e = 0.609506, i = 8.775504, peri = 82.751300.

### DIFF
--- a/wmpl/Rebound/REBOUND.py
+++ b/wmpl/Rebound/REBOUND.py
@@ -10,6 +10,7 @@ from that shell. Installing only 'rebound' is not enough; 'reboundx' must import
 import os
 import re
 import sys
+import json
 import time
 import warnings
 import concurrent.futures
@@ -233,7 +234,15 @@ def findEarthDepartureIndex(sim_outputs, n_hill=3.0):
 
 
 def detectCloseEncounters(sim_outputs, n_hill=3.0):
-    """ Detect close encounters between the integrated object and the planets (and the Moon).
+    """ Detect close encounters between the integrated object and the planets (and the Moon), by
+    scanning the sampled output.
+
+    DEPRECATED for new code, because scanning the output is sampling-limited: near the Earth the
+    object can cross a large fraction of the Moon's detection sphere between two output samples, so a
+    lunar encounter can be missed outright or its minimum distance overestimated (by ~60000 km at
+    n_outputs = 100 on a real trajectory). Prefer encountersFromMinDistances, which uses the exact
+    minima recorded at every internal integrator timestep. This function is kept for callers that
+    only have sampled output to work from.
 
     A close encounter is flagged when the minimum object-body distance drops below n_hill times
     the body's Hill-sphere radius (see HILL_RADII_AU). The Hill sphere is the standard criterion
@@ -1233,7 +1242,10 @@ def reboundSimulate(
         sim_days: [float] Length of integration in days, default is 60 days.
         n_outputs: [int] Number of outputs (samples along the simulation), default is 500.
         obj_name: [str] Name of the object that's being integrated, default is "obj".
-        obj_mass: [float] Mass of object in solar masses if asteroid or larger object, default is 0.0.
+        obj_mass: [float] Accepted for backwards compatibility but ignored: the integrated object is
+            always treated as a massless test particle, so that it cannot perturb the planets and the
+            Monte Carlo realizations cannot perturb each other. Give the object a mass only by
+            editing _integrateParticles, and note that N_active would have to change with it.
         mc_runs: [int] Number of Monte Carlo simulations to run, default is 100.
         reference_frame: [str] Reference frame to use for the state vector. Options:
             - "heliocentric" (default)
@@ -1611,7 +1623,11 @@ if __name__ == "__main__":
         print("Running the simulation in geocentric reference frame.")
 
     
-    # Set semi-major axis and periapsis units depending on the reference frame
+    # Set semi-major axis and periapsis units depending on the reference frame. Note that this
+    # multiplier applies only to the orbital elements: the object-body distances, the close-encounter
+    # distances and the divergence figures are always reported in AU (with km given alongside where
+    # it helps), because they are distances between bodies rather than orbit sizes and so do not
+    # depend on the frame the elements are expressed in.
     a_units = "AU"
     q_units = "AU"
     dist_unit_multiplier = 1.0  # Default is AU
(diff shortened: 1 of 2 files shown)


dvida:
## Rigorous self-review — 4 defects found, reproduced, and fixed

I wrote this PR, so the review was built to falsify its claims. It found four real defects. Each was reproduced *before* fixing; fixes pushed in 3aa5f21.

### Confirmed defects (fixed)

**D1 — one truncated MC realization destroyed the divergence estimate for the entire ensemble.** `estimateLyapunovFromMC` took `min()` of sample counts across the nominal run *and every realization*, so a single realization ended early by an impact — a possibility this very PR introduced — collapsed the fit. **Reproduced: one 3-sample realization out of 20 reduced the whole result to `None`.** The two new features actively broke each other. Now each realization contributes over its own length; truncated realizations are counted and reported.

**D2 — plot crashed on length mismatch.** Every MC series was plotted against the *nominal* time axis. **Reproduced: `ValueError: x and y must have same first dimension, but have shapes (200,) and (3,)`.** Each realization now uses its own time axis.

**D3 — `HILL_RADII_AU` was internally inconsistent.** Recomputing from the code's own masses disagreed by up to **5.7%** (Luna +5.7%, Mercury +4.1%, Saturn +3.7%, Uranus +3.3%); Luna matched the circular form while others matched neither. Regenerated from one documented convention, `r_H = a·(m/3M☉)^⅓`, using the same GM table as the simulation masses. Chose the larger (no eccentricity factor) form deliberately, since these define an encounter *screening* threshold. Earth R_Hill 0.009810 → 0.010004 AU; departure gate 0.029430 → 0.030012 AU.

**D4 — the Lyapunov classifier was overconfident.** On a 3000-day run it reported **"Lyapunov time = 980.79 days" from R²(exp) = 0.603 vs R²(lin) = 0.507** — two equally poor fits separated by 0.096 — and asserted "the integration loses predictive value beyond this timescale." The same run detected **no close encounters at all**, so there was no mechanism for genuine chaos; the growth was almost certainly along-track phase drift. Acceptance now requires R² ≥ 0.90 absolute plus a 0.05 margin, and a new `saturated` state rejects the estimate once the ensemble exceeds 10% of its heliocentric distance, where the linearised picture behind a Lyapunov exponent breaks down. The same case now correctly reports that neither law fits and claims **no** timescale.

### Claims that survived scrutiny
- Masses match reference values to ≤3e-11; Sun exactly 1.0.
- "Moon radius was 7% too small" → **6.9%** (1617.9 vs 1737.4 km). Accurate.
- The Moon-gating geometry: apogee 0.00271 AU, 3 R_Hill,Luna 0.001233 AU (184,454 km), lunar encounter requires Earth distance < 0.00394 AU vs a 0.0294 AU gate — all exact. The argument holds.
- Sampling-error claims re-derived: 464 km at `n_outputs=500`, **59,524 km** at `n_outputs=100`.
- Particle indices are stable: nothing calls `sim.remove()` and `collision_resolve = "halt"` leaves `N` unchanged (verified directly).
- Two-value return signature preserved by default.

### Notable gap closed
The `"exponential"` branch **had never executed in any run**. It is now exercised against synthetic ensembles and recovers true Lyapunov times of 20, 50 and 150 days **exactly**, while still classifying pure linear growth correctly.

### Documented limitations (not fixed)
- The exponential-vs-linear decision compares R² of fits to *different dependent variables* (`ln δ` vs `δ`), which is not strictly rigorous. The strict thresholds and the saturation guard mitigate it; a residual-based or information-criterion test would be the principled replacement.
- Separation uses position only, not full phase space.

### Environment note (unrelated to this PR)
A stale duplicate `wmpl` exists in `site-packages`, so any script run from outside the repo imports that copy rather than the working tree — worth cleaning up to avoid confusing results.

Reference case (60 d, `--mc 100`) nominal elements unchanged: a = 1.454858, e = 0.609506, i = 8.775504, peri = 82.751300.

PR #94, opened 2026-09-16
Floor the RC2001 deceleration correction instead of tapering it to zero

AstroEloy:
## Why

Follow-up to 096398e, which correctly identified that the ReVelle & Ceplecha (2001)
deceleration correction

    f(dv) = 0.26*ln(dv) + 0.0042*ln(dv)^3,   dv = (v_init - vel) in km/s

diverges to -inf as dv -> 0 and silently returned NaN for vel > v_init. Both problems are
real and the bound was needed. This PR keeps the bound but changes its shape.

The issue with the linear taper is that it runs against the correction's own trend.
`f` is strictly increasing in dv everywhere:

    df/dL = 0.26 + 0.0126*L^2 > 0   for all L = ln(dv)

The taper, `(dv/dv_min)*f(dv_min)`, is *decreasing* in dv. Splicing them puts a spurious
local minimum at exactly dv = dv_min:

| dv (km/s) | 0 | 0.025 | 0.05 | **0.1** | 0.3 | 1.0 | 5.0 |
|---|---|---|---|---|---|---|---|
| tau / tau_uncorrected | 1.00 | 0.85 | 0.72 | **0.52** | 0.73 | 1.00 | 1.55 |

tau dips to 52% and climbs back on *both* sides, so the mass integrand carries a local
bump of x1.92 at dv = 0.1 km/s. The splice is also a sharp V: the derivative jumps from
-6.5 to +3.27.

Two further consequences:

- The taper moves *against* the direction the empirical fit indicates. RC2001 says tau
  falls as dv falls; the taper returns tau to its fully uncorrected value at dv -> 0.
- Combined with 06b76b9 (v_init = median of the first 25% of points), the start of a noisy
  trajectory now crosses three regimes in a fraction of a second: dv <= 0 (no correction),
  the taper, then f.

## What this changes

Floor the *argument* rather than reshaping the output:

    if dv < dv_min:
        dv = dv_min
    decel = 0.26*log(dv) + 0.0042*log(dv)**3

This holds the fit at the edge of its calibrated range instead of extrapolating a new
functional form into a region where RC2001 was never constrained — the standard way to
handle extrapolation beyond a fit's validity. It also removes the separate `dv <= 0`
branch: `dv_eff >= dv_min > 0` makes the NaN unreachable by construction, so the guard is
no longer a special case. Net -1 line.

The trade-off, stated plainly: the floor asserts tau x0.52 where no deceleration is
measurable, whereas the taper asserted no correction at all. Neither is measured — RC2001
has no calibration at dv -> 0 — but the floor does not invent a shape, and it keeps the
correction monotonic.

## Measured impact on photometric mass

Synthetic light curves, 100 Hz, v_init = 20 km/s, RC2001 Type II:

| case | change |
|---|---|
| dv_total = 10 km/s, late peak | +0.3% |
| dv_total = 3 km/s, late peak (typical) | +1.0% |
| dv_total = 0.5 km/s, late peak | +2.0% |
| dv_total = 0.1 km/s, late peak | +22% |
| dv_total = 3 km/s, early peak | +18% |
| dv_total = 0.5 km/s, early peak | +38% |

Negligible for well-decelerating meteors with late peaks; the large differences are exactly
the weakly-decelerating and early-flare cases where the taper distorted the integrand most.

## Verification

Against the compiled extension, not a reimplementation:

- Monotonic non-decreasing across the full physical sweep (min delta = -3.3e-16, FP noise)
- No NaN/Inf anywhere, including vel > v_init
- Floor exact at exp(f(0.1)) = 0.522074; neutral point exact at dv = 1 km/s (f = 1.000000)
- `wmpl/MetSim/Tests/test_MetSimErosion.py` passes

Docstring and the `calcMass()` WARNING text in `Physics.py` updated, since both described
the taper.

🤖 Generated with [Claude Code]


dvida:
## Review

The change is correct and I would merge it. I checked the numbers against compiled builds of both versions of `MetSimErosionCyTools.pyx` (this branch and master), not against a Python reimplementation, and pushed one commit on top: `f9e7ab1`, regression tests.

### Verified

At 20 km/s, 1 kg, RC2001 Type II, τ relative to the uncorrected value:

| dv (km/s) | master (taper) | this PR (floor) |
| --- | --- | --- |
| ≤ 0 | 1.000 | 0.522 |
| 0.05 | 0.723 | 0.522 |
| **0.10** | **0.522** | 0.522 |
| 0.30 | 0.726 | 0.726 |
| 1.0 | 1.000 | 1.000 |
| 5.0 | 1.546 | 1.546 |

- Master's τ has a local minimum exactly at dv = 0.1 km/s (smallest step −1.4e-5 over a 200k-point sweep), as the description says; the PR's τ is non-decreasing over −5..40 km/s, finite everywhere including vel > v_init, and the floor equals exp(f(0.1)) = 0.522074 to machine precision. dv = 1 km/s is neutral in both.
- Types I and III are floored the same way; the non-RC2001 models ignore `v_init`.
- Photometric mass with `calcMass()` on synthetic 100 Hz light curves at 20 km/s, master → PR: +0.5 % (dv = 10 km/s, late peak), +1.5 % (3 km/s, late), +3 % (0.5 km/s, late), +21 % (0.1 km/s, late), +10 % (3 km/s, early peak), +34 % (0.5 km/s, early). With `v_init = -1` (median of the first quarter) the same cases give +1 % to +25 %. Same pattern and size as the table in the description.

The formula's coefficients themselves are pre-existing and not touched here. The trade-off is stated honestly in the description: neither the floor nor the taper is measured below 0.1 km/s, but only the floor keeps the correction monotonic, and it avoids the three-regime start that `v_init = -1` produces with the taper.

### Added

`f9e7ab1` — `wmpl/MetSim/Tests/test_LuminousEfficiency.py`, 10 tests: finite and non-decreasing in dv for Types I–III, exact floor below 0.1 km/s and for vel ≥ v_init, neutral at 1 km/s, other models ignore `v_init`. Run against the compiled master version, the six floor/monotonicity tests fail and the four shared-property tests pass, so the suite discriminates the two bounds rather than merely passing.

PR #95, opened 2026-09-17
DynamicMassFit final azim elev

AstroEloy:
## Summary
- `DynamicMassFit.computeFragEndParams()` returns the final lat/lon/height of the propagated
  fragment. It now also returns the apparent ground-fixed radiant azimuth and elevation at that
  point, which dark flight needs as input. Both are printed in the CLI summary for the three mass
  cases.
- `SampleTrajectoryPositions.sampleTrajectory()` precessed the derotated radiant from J2000 to the
  epoch of date before converting it to azimuth/elevation. The radiant is already in the epoch of
  date, so precession was applied twice. That step and its imports are removed.

## Why the coordinates are already in the epoch of date
- Station positions come from `geo2Cartesian()` with its default `precess_j2000=False`.
- Lines of sight come from `altAz2RADec()`: RA = LST - HA, with apparent sidereal time.
- `raDec2AltAz()` uses LST directly and expects epoch-of-date RA/Dec.
- `Orbit.calcOrbit()` performs the same derotation and conversion without precession, and labels
  the result "apparent ground-fixed, epoch of date".
- `SampleTrajectoryPositions` itself prints "The horizontal coordinates are apparent above a fixed
  ground, not topocentric in J2000!".

## Why at the final point
The local horizon rotates along the path by ~0.009 deg per km of ground track, so
`traj.orbit.azimuth/elevation_apparent_norot`, evaluated at the reference point, is not valid at
the end. The difference is tenths of a degree for final points tens of km downrange, and several
degrees for shallow trajectories propagated hundreds of km.

## Method
`DynamicMassFit` uses the same derotation as `SampleTrajectoryPositions` and `Orbit.calcOrbit()`,
evaluated at the final point and time. It uses the Earth rotation velocity and east direction at
that point, the radiant from `traj.radiant_eci_mini`, and the final JD and lat/lon for the
conversion, without precession.

The derotation uses the average speed `v_avg_norot`, as `SampleTrajectoryPositions` does. Drag acts
on the air-relative velocity, so the air-relative direction is conserved and the path is straight in
the Earth-fixed frame. Recovering that direction from a straight-line ECI fit requires an average
speed over the fitted segment. This was tested on synthetic fireballs with a known air-relative
direction, fitted as the solver does. Average speeds recover the direction to <= 0.02 deg for a slow,
strongly decelerating fireball, and to <= 0.005 deg for medium and fast ones. `v_init` or the
terminal ~3 km/s speed would be off by ~1 deg in the slow case.

## Validation
The reference was computed independently of wmpl with pyerfa (IAU SOFA), from kinematics only: the
inertial velocity rotated to Earth-fixed with the Greenwich apparent sidereal time, minus w x r,
projected onto the local geodetic ENU basis, with no RA/Dec conversions and no derotation formula.
It agrees with astropy to 0.4" and with a finite difference of Earth-fixed positions to 3e-6 deg.

On several real trajectory solutions (slow and fast fireballs, both hemispheres):
- Stored station ECI positions match epoch-of-date coordinates to <= 0.13". In J2000 they would be
  0.14-0.34 deg (16-37 km) away, so the trajectory frame is epoch of date.
- `traj.orbit.azimuth/elevation_apparent_norot` matches the reference to < 5e-6 deg.
- `DynamicMassFit`: the returned final azimuth/elevation matches the reference to <= 1e-4 deg.
- `SampleTrajectoryPositions`: before this PR it is off by 0.13-0.54 deg in azimuth or elevation at
  every sample. That output matches reading the radiant as J2000 (to ~0.01 deg, i.e. nutation),
  not as epoch of date. After this PR it matches the reference to <= 1e-4 deg.

The removed error has the signature of precession. For a fixed ground-fixed direction evaluated
between 1980 and 2050, it vanishes at J2000, changes sign across it, and grows at 48.65"/yr and
35.45"/yr for two test directions. General precession predicts 48.64"/yr and 35.51"/yr (50.29"/yr
times the sine of the angle to the ecliptic pole). Without the step, the error stays <= 4e-5 deg over
the whole range.

`TrajectoryKML` only uses the sampled lat/lon/height, so it is unaffected. Both CLIs run end to end.

🤖 Generated with [Claude Code]


dvida:
## Review

Both changes are correct. I verified the frame claim independently of wmpl, checked the covariance change against a bootstrap, found two small defects, and pushed one commit on top: `2e46bf7`.

### The double precession is real

Reference built with astropy only: the ECI (true equator and equinox of date) velocity transformed to ITRS, minus ω × r, projected onto the geodetic ENU basis — no RA/Dec conversion and no derotation formula. Against it, on three real solutions:

| trajectory | epoch | `SampleTrajectoryPositions` on master | on this branch |
| --- | --- | --- | --- |
| 20191023 (repo example) | 2019.8 | 0.54° in azimuth | 1.3e-3° |
| 20240530 | 2024.4 | 0.39° az / 0.18° el | 1.3e-4° |
| 20201005 | 2020.8 | 0.24° in elevation | 6.7e-4° |

The same reference reproduces `traj.orbit.azimuth/elevation_apparent_norot` to 5e-5°, so the solver's frame is epoch of date and Orbit's construction is right — the sampling code was the odd one out. The other `equatorialCoordPrecession(J2000_JD, ...)` calls in the repo all act on inputs that really are J2000 (ShowerSim, CMNFormats, `cartesian2Geo(precess_j2000=True)`, Orbit's optional J2000 input), so this was the only instance.

`DynamicMassFit` end to end on 20240530: final azimuth/elevation 149.108° / 51.721° at the final point, against 149.173° / 51.782° at the reference point, i.e. the horizon rotation and gravity turn the description promises are the right size.

### The covariance fix holds up

Residual bootstrap (1000 resamples) of the branch's own estimator on the 20240530 velocity window, 12 points: σ(slope) = 486 m/s². The branch's formula gives 478 (ratio 0.98); master's SVD of `res.jac` gives 614 (1.26). On a single robust fit with the default `f_scale` the old formula was off by 10× (5624 vs 515). The change is right. Note the pairs bootstrap gives 1374 on the same data — the ±2σ interval is still a deceleration-only, fixed-design estimate, which the new report line now says.

### Found and fixed (`2e46bf7`)

1. **`--maxvel` was parsed and never used.** The velocity filter hardcoded 73 km/s regardless, so the option did nothing; this branch reworded its help ("Default and upper limit is 73 km/s") without noticing. The filter now uses `min(max_vel, 73000)`.
2. **`np.linalg.inv(J.T @ J)` raised a raw `LinAlgError`** when all points share one time (hit in a pairs bootstrap). The old SVD path raised a clear `RuntimeError`; that case is refused again with a clear message.
3. **The derotation block was now written three times** (Orbit, SampleTrajectoryPositions, DynamicMassFit). Extracted into `TrajConversions.derotatedRadiantAltAz(v_eci, eci_pos, jd, lat, lon)`, used by the two files in this PR; Orbit keeps its copy. CLI output is unchanged to the last printed digit and the sampled azimuth/elevation match this branch's earlier values.

Also: wmpl-style docstrings with `Arguments:`/`Return:` for `pointOnTrajectory` and `computeFragEndParams`, and `wmpl/Utils/Tests/test_DynamicMassFit.py` (7 tests on the repo's example trajectory: helper vs the solver at the reference point to 1e-6°, helper vs the astropy reference along the path to 2e-3° — skipped without astropy — sampled azimuth within 0.05° of the solver's where master was 0.54° off, `pointOnTrajectory` drop perpendicular and downward, covariance = σ²(JᵀJ)⁻¹, zero time span refused, outlier rejection).

### Two things for the author

- **The description is stale on one point.** It says the derotation uses `v_avg_norot` "as SampleTrajectoryPositions does"; the code uses `v_init` since `9826b67`, with the end-to-end evidence in that commit message (0.13° vs 0.18°). The code is the better choice; the description should say so.
- **17 commits with a revert pair** (`7808ca1` / `99681c9`) for a +150-line change — worth a **squash merge**.

PR #98, opened 2026-09-18
Support REBOUND 5 and fix reboundx builds against the wrong REBOUND

AstroEloy:
## Summary

Makes `wmpl/Rebound/REBOUND.py` work with both REBOUND 4 and REBOUND 5, and fixes a REBOUNDx install problem that already breaks fresh installs of the pinned 4.3.0.

### Problem 1: REBOUNDx is compiled against the wrong REBOUND

`reboundx` publishes no wheels, so `pip` always compiles it. By default it compiles in an isolated build environment against the newest `rebound` on PyPI, not the installed one. When the versions differ, the C `reb_simulation` structs differ too. With `rebound==4.3.0`, for example, `reboundx` writes `sim->extras` at byte 2272 of a 2128-byte struct, which overwrites heap memory. Every `.params[...]` then fails with:

```
AttributeError: Need to attach reboundx.Extras instance to simulation before setting params.
```

Upgrading `rebound` after `reboundx` was compiled causes a related failure: `import reboundx` raises `OSError` (`Symbol not found`). This escaped the module's `except ImportError`, so importing `REBOUND.py` crashed.

Compiling with `pip install --no-build-isolation reboundx` avoids both.

### Problem 2: REBOUND 5 API changes

conda-forge now serves `rebound` 5.x, so following the README installs it. Three things break:

- `sim.add(..., hash=...)` was removed in favour of `name=`.
- Bodies queried from Horizons cannot be named through `sim.add` in v5, because the Horizons query takes `name` as the body to look up.
- `sim.heartbeat = func` raises `AttributeError: '_hb'` for any function in REBOUND 5.1.1. This is an upstream bug: `_hb` is missing from `Simulation.__slots__`.

## Changes

- **`REBOUND.py`**
  - `_addNamedParticle` uses `hash=` on v4 and `name=` on v5, and names Horizons bodies after adding them.
  - `_setHeartbeat` falls back to setting the C function pointer directly, only when the setter fails with the `_hb` error.
  - `_checkReboundxAttached` raises a clear error with the reinstall command if `reboundx` failed to attach.
  - A `reboundx` that fails to load (`OSError`) is now reported like a missing dependency, through the existing deferred error message, instead of breaking the import.
  - The install help printed on a failed import recommends `--no-build-isolation` and explains how to recompile an existing `reboundx`.
- **README**: installs `reboundx` with `--no-build-isolation`, explains how to repair an existing install (and how to keep an older `rebound`), and the verification one-liner now detects a mis-compiled `reboundx`.
- **Tests**
  - New `test_REBOUND_integration.py`: unit tests for the helpers, which run without REBOUND, plus an end-to-end integration of the example trajectory against reference values. The reference includes the lunar closest approach, so the test also checks that the heartbeat runs.
  - The integration tests are skipped when REBOUND, the DE430 ephemeris or the example trajectory are unavailable.
  - `test_REBOUND.py` gets a test for the `OSError` case.
  - The existing tests never integrated, so they passed even on a broken install.

## Testing

| rebound + reboundx | Tests | Notes |
|---|---|---|
| 4.3.0 + 4.3.0, built against it | 34/34 | Reference |
| 4.6.0 + 4.6.2 (latest 4.x) | 34/34 | |
| 5.0.1 + 5.0.1 (conda-forge) | 34/34 | |
| 5.0.1 + 5.1.0 (the README route) | 34/34 | |
| 5.1.1 + 5.1.0 (latest) | 34/34 | |
| 4.3.0 + reboundx built against a newer rebound | 32/34 | The two integration tests fail with the new, explicit error |
| 5.0.1 + stale reboundx 4.3.0 | 31 passed, 3 skipped | Module imports cleanly; the CLI reports the load error and how to fix it |

- **CLI on 4.3.0 and 5.1.1** (default, `--forward --beta`, `--horizons`, `--geocentric`): the JSON output agrees to ≤1e-12 for orbital elements and encounter distances.
- **Monte Carlo** (8 clones with a synthetic covariance): serial and 4-process runs are bitwise identical on each version, and 4.3.0 and 5.1.1 agree to 4e-12.
- **README repair routes**: both (pin `reboundx==4.3.0`, or upgrade both to 5.x) restore 34/34 on a broken environment.

## Not changed

`requirements.txt` still pins `rebound==4.3.0` / `reboundx==4.3.0`. While that pin stays, `pip install -r requirements.txt` builds `reboundx` in isolation and can still produce the mismatch from problem 1; with this PR it is at least reported clearly. Moving the pin to 5.x, or dropping it, is left for discussion.


dvida:
## Review pass: 1 commit pushed to this branch

Reviewed for correctness, style, comment density, tests and documentation. **The diagnosis and both fixes are right**, and this is the PR in the stack that catches the class of failure the others cannot: a `reboundx` compiled against the wrong `rebound` silently corrupts heap memory and then fails with an error message that points nowhere. 34/34 tests pass as submitted on rebound 4.5.0 / reboundx 4.6.0.

I can only test one version pair here, so the 4.3.0 / 5.0.1 / 5.1.1 matrix in the description is taken on trust. Everything below is from reading the code and from the merged stack.

### First, a retraction

I told the maintainer earlier that this PR's new `test_REBOUND_integration.py` "may be largely inert on any machine without the example trajectory, including CI", because it skipped with *"Example trajectory not found"* in my sandbox. **That was my harness, not the repo.** `wmpl/Dynesty/examples/20191023_091225/20191023_091225_trajectory.pickle` is tracked in git, and run from a real checkout the file gives 34/34 with nothing skipped. The end-to-end test does run.

### `33f5882`

**The REBOUNDx check fires too late.** `_checkReboundxAttached` is only called inside `_integrateParticles`, which runs in the worker processes. On a mis-compiled install the user therefore sits through the whole ephemeris setup and then gets the same `RuntimeError` once per worker — for a 200-clone run on 8 cores, eight copies of a 6-line message, after the slow part. Since the entire point of the function is a clear early error, `checkReboundxUsable()` now makes the same check on a throwaway simulation and `reboundSimulate` calls it immediately, next to the existing `REBOUND_FOUND` guard. (#100 added a `checkIntegratorAvailable` call in the same place, for the same reason.)

**`int(rb.__version__.split(".")[0])` ran on every particle add** and raises `ValueError` on any version string that does not start with a digit. It is now `_reboundMajorVersion()`, parsed in one place, falling back to 5 — the safe assumption, since the naming keyword only changed going forward.

**Commented the `_setHeartbeat` fallback.** It writes `sim._heartbeat` with `rb.simulation.AFF(func)`, i.e. reaches into REBOUND's internals, which is the right call for an upstream bug but should say out loud what it is doing and why the other `AttributeError` is re-raised.

### Tests, 34 → 41

**`_checkReboundxAttached` had no negative test.** The only case asserted that it *passes* on a working install, so the failure path — and the `--force-reinstall` command inside it, which is the actual deliverable of this PR — was never exercised. A stand-in object with a falsy `extras` now covers it, including that the message carries the repair command.

Also: `checkReboundxUsable()` on a working install, and `_reboundMajorVersion()` over ordinary, short and non-numeric version strings.

Mutation-checked: removing the version fallback or neutering the attach check fails 3 tests.

### Documentation

`wmpl/Docs/REBOUND.md` (added in #100) said `pip install reboundx` — precisely the command that produces the broken install this PR exists to detect. Updated on **#100's branch** (`686c975`), since that is where the file lives: `--no-build-isolation` and why it is not optional, the verification one-liner that also checks REBOUNDx attached, the repair route, both resulting error messages in Troubleshooting, and a note that the code supports REBOUND 4 and 5.

### One open question, from your own "Not changed" section

`requirements.txt` still pins `rebound==4.3.0` / `reboundx==4.3.0`, so `pip install -r requirements.txt` still builds `reboundx` in isolation and can still produce the mismatch — now reported clearly rather than silently, which is a real improvement, but a fresh `pip`-route install remains broken by default. Worth resolving before this merges, since the README now tells people to use `--no-build-isolation` while the requirements file quietly does not.

---

## The stack: #98 → #99 → #100, merged and tested for real

I have now reviewed all three. I merged them locally in that order, in a real checkout so nothing skips:

**Result: 83 passed, 0 failed, 0 skipped.**

An end-to-end CLI run on the merged tree (`--integrator trace --dt 0.5 --compute_megno`, 300 years backward, real trajectory) exercises all three PRs at once and comes out right:

```
  Integrator   : TRACE, dt = 0.5 d, from t = -36.537 d (IAS15 before, while leaving the Earth)
  Energy drift : 1.51e-10 (relative)
  Close encounters (< 3 Hill radii), in the order they happened:
    Jupiter      0.822781 AU  at t =  -12492.351 d   (2.32 R_Hill)
    Jupiter      0.567932 AU  at t =  -47130.364 d   (1.60 R_Hill)
    Jupiter      0.860763 AU  at t =  -81726.023 d   (2.42 R_Hill)
    Jupiter      0.451237 AU  at t =  -94757.545 d   (1.27 R_Hill)
    Jupiter      1.056789 AU  at t = -107685.689 d   (2.97 R_Hill)
```

Those times being **negative** is the point: a backward TRACE run integrates in reversed time, so without follow-up 1 below they would all come out positive.

**Textual conflicts**, one hunk per step, all trivial except two that must take a side:

| Step | Conflict | Resolution |
|---|---|---|
| +#99 | `heartbeat_ref = _setHeartbeat(sim, heartbeat)` vs #99's refinement block ending `sim.heartbeat = heartbeat` | Keep #99's block, end it with this PR's `_setHeartbeat` line |
| +#100 | `_checkReboundxAttached(sim)` / `sim.dt = …` | Keep `_checkReboundxAttached(sim)` and **#100's** `sim.dt = … integrate_forward`. Keeping both `sim.dt` lines runs correctly by accident — the second assignment wins — but leaves a dead line |
| +#100 | `cloneEncounterSummary` vs `whfastEncounterWarning` | Keep both |
| +#100 | `encounters = sorted(…)` vs #100's integrator/WHFast/MEGNO block | Keep #100's block but **drop its** `encounters = encountersFromMinDistances(...)`. Naive keep-both leaves #99's list silently overwritten |
| +#100 (tests) | Adjacent sections in `test_REBOUND_numerics.py` | Keep both |

**Semantic follow-ups no merge tool catches** — five, all now verified against a real merge:

1. **#99's encounter times need #100's `time_sign`.** Otherwise a backward TRACE run reports positive encounter times.
2. **`computeMegno` must use this PR's `_setHeartbeat`.** #100 assigns `sim.heartbeat` directly, which fails on REBOUND 5.1.1.
3. **`whfastEncounterWarning` must read `diag["encounters"]`** rather than calling `encountersFromMinDistances`. Not in any description — that function was added during the #100 review.
4. **Its test fixtures need an `"encounters"` list** once item 3 lands. Not in any description; without it the merged tree has 2 failures.
5. **This PR's lunar closest-approach expectation.** `test_REBOUND_integration.py` fixes it at `-0.0426370612` d, the step-end value. Under #99's within-step refinement it becomes **`-0.0426156724`** d, which I can now confirm from a real run — #99's description estimated `-0.0426157`. The distance assertion still passes unchanged at `rel=1e-7`, consistent with #99's finding that the Moon was already well resolved.

Items 1–5 are deliberately **not** applied to any of the three branches: each PR is self-consistent and passes on its own, and item 5 in particular would break this PR standalone. They belong to whoever merges.

PR #99, opened 2026-09-18
REBOUND: refine close-encounter minima between steps and list every encounter

AstroEloy:
## Problem

`_integrateParticles` records each body's closest approach at the end of every IAS15 step, and the docstrings and the results file call that value exact. It is not exact. Away from the Earth, IAS15 takes steps of 1–2 days, so a meteoroid can move millions of km past a planet within one step. The step-end distance can then be far above the true minimum, and an encounter inside 3 Hill radii can be missed.

Nothing downstream corrects for this. `encountersFromMinDistances` only applies the threshold, and the value goes unchanged into the terminal summary, the results file, the JSON and the plot.

## Measurements

The reference is a dense re-integration (a few seconds between outputs) of 1196 synthetic flybys: 600 lunar and 600 planetary, with relative speeds of 5–70 km/s. The planetary flybys have impact parameters up to 3.5 R_Hill.

| Setting | Current code | This PR |
|---|---|---|
| Long output spacing (e.g. 100 yr / 500 outputs) | Minimum overestimated by up to 215 % (Mercury) and 45–50 % (Venus, Earth, Mars). Encounters < 3 R_Hill missed: Mercury 25/104, Venus 6/108, Mars 4/107, Earth 2/106. Closest-approach time off by up to ~18 h | 0 missed, distance within 1 km, time within 0.2 min |
| CLI default (60 d / 500 outputs) | Nothing missed. Errors up to 24 % (Mercury), 8 % (Mars), 4 % (Venus). Time off by up to 43 min | Same as above |
| Moon (either setting) | Already fine (< 1.5 %): steps near the Earth are short | Same as above |

- **Missed encounters:** every one had a true minimum between 2 and 3 R_Hill.
- **End-to-end check:** I ran `reboundSimulate` → `encountersFromMinDistances`, the same path the CLI uses, over 100 years. On the current code, flybys of Mercury (2.04 R_Hill), Venus (2.27) and Mars (2.27) do not appear. With this change they do, at the dense-reference distance and time.

## Fix

The heartbeat keeps the previous step's relative state for each body. When the approach rate r·v changes sign across a step, the distance has a minimum inside that step. The new `hermiteClosestApproach` then finds it on the cubic Hermite interpolant of the relative position and velocity at the two step ends.

- **Weakly deflected flybys:** the interpolant reproduces straight-line motion exactly, so the result is accurate even when one step spans the whole encounter.
- **Strongly bent flybys:** the interpolant would underestimate the minimum if a step spanned the whole flyby. IAS15 shortens its steps in that regime, though. A stress test of slow, deep flybys of every planet (hyperbolic eccentricity down to ~1) stayed within 3e-4 of the dense reference.

## Every encounter, not only the deepest per body

The diagnostics kept one minimum per body, so repeated encounters with the same planet collapsed into one: an earlier passage at 2–3 R_Hill disappeared behind a later, deeper one. The heartbeat already detects every local minimum of each object–body distance and refines it on the interpolant. Each refined minimum inside `n_hill` Hill radii is now also appended to a per-particle `"encounters"` list (body, time, distance, R_Hill, closeness), in the same format as `encountersFromMinDistances`. This adds no cost, because those steps were already being refined.

- **CLI report:** the terminal, the results file, the JSON and the plot markers list the encounters in the order they happened.
- **Monte Carlo summary:** per body, it gives the clones with at least one encounter, the total number of encounters and the closest approach. The JSON also stores each clone's list.
- **Unchanged:** "Closest approach to each body" and `encountersFromMinDistances`.
- **Example** (synthetic Mercury flyby, 100 yr, through the actual CLI): Mercury at t = 1.0 d (2.04 R_Hill) and at t = 11172 d (2.59 R_Hill). Before this PR neither was reported: the first because of the step sampling, the second because only one minimum per body was kept. With 24 clones: Mercury 8/24 clones, 10 encounters in total.

## Cost

- **Integration unchanged:** orbits and all outputs are bitwise identical. Only `min_dist_au` and `min_time_days` change, and never upwards.
- **Runtime:** about 3.5 % more wall time (5.02 → 5.20 s for a 100-year run). The per-step work uses plain floats; building numpy arrays there cost about 13 %.
- **Dependencies:** none added.

## Tests

- **Hermite numerics** (`test_REBOUND_numerics.py`): straight-line flyby inside one long step, backward step, step with no interior minimum, and Keplerian hyperbolic flybys at three eccentricities.
- **Integration regression test** (`test_REBOUND_encounters.py`): a synthetic Mercury flyby on circular planet orbits, so no ephemeris file is needed. The minimum with natural IAS15 steps must match a dense re-integration, in both directions. It fails on the current code (1 % forward, 12 % backward overestimate) and is skipped without rebound/reboundx.
- **Encounter list** (`test_REBOUND_encounters.py`): an object with Mercury's period on an inclined orbit meets Mercury at alternating nodes every ~44 days. All three passages in 100 days must be listed in order and match a dense re-integration, and the deepest must equal `min_dist_au`.
- All 32 Rebound tests pass on rebound 4.3.0 and 4.6.0.

## Interaction with #98

The two PRs are independent, but whichever is merged second needs two small follow-ups:

1. **Conflict:** there is a one-line conflict at the end of the heartbeat. #98 replaces `sim.heartbeat = heartbeat` with `heartbeat_ref = _setHeartbeat(sim, heartbeat)`; keep both the refinement block and #98's line.
2. **Test value:** #98's `test_REBOUND_integration.py` fixes the lunar closest-approach time at the step-end value, -0.0426370612 d. With the refinement it becomes -0.0426157 d, which is the dense-reference value, so that expected value has to be updated.

With both applied, 42/43 Rebound tests pass on rebound 5.1.1 and 4.3.0; the only failure is that expected value.


dvida:
## Review pass: 2 commits pushed to this branch

Reviewed for correctness, style, performance, comment density, tests and documentation. **The numerics are right.** The Hermite refinement and the `r·v` sign test are correct in both time directions — `(h*rv_prev < 0) and (h*rv > 0)` is the right invariant for `h < 0` too, which is easy to get wrong and this does not. 32/32 tests passed as submitted on rebound 4.5.0 / reboundx 4.6.0. The two commits below are polish, not repairs.

### Independent confirmation of the two central claims

**Performance.** On the real 10-body system (100 years backward, 2000 outputs, a real trajectory pickle), best of two runs each:

| | wall time |
|---|---|
| master | 22.63 s |
| this PR | 23.10 s (**+2.0 %**) |
| this PR + my commits | 23.08 s |

So the stated ~3.5 % is if anything conservative, and the refactor below costs nothing. On a 4-body synthetic task the overhead is +8 %, which is consistent rather than contradictory — with fewer bodies the force evaluation is cheaper, so the fixed per-step heartbeat cost is a larger fraction of it.

**The fix does what it says.** On a real trajectory integrated 300 years backward, the encounter list now reads:

```
  Close encounters (< 3 Hill radii), in the order they happened:
    Jupiter      0.822781 AU ( 123086329.4 km)  at t = -12492.351 d   (2.32 R_Hill)
    Jupiter      0.567932 AU (  84961451.9 km)  at t = -47130.364 d   (1.60 R_Hill)
    Jupiter      0.860776 AU ( 128770285.2 km)  at t = -81726.021 d   (2.42 R_Hill)
    Jupiter      0.451241 AU (  67504680.6 km)  at t = -94757.544 d   (1.27 R_Hill)
    Jupiter      1.056882 AU ( 158107256.1 km)  at t = -107685.680 d   (2.97 R_Hill)
```

Before this PR that was a single line. Four of the five passages were invisible.

### `3c7e545` — encounter record, plot labels, docstrings

- **That same example exposed a plot problem.** Every passage got a red star *and* a text label, so those five encounters produced five overlapping "Jupiter" labels on one panel. Every passage still gets a star; only the deepest per body is now named.
- **`encountersFromMinDistances` has no call site left in `__main__`.** Both uses became direct `diag["encounters"]` reads, which is right, but it now reads as dead code. Its docstring says what it is still for: building the list for a caller that only has the closest-approach dictionaries. The paragraph explaining this was also sitting between `Arguments:` and `Return:`, where wmpl docstrings do not put prose.
- **`_encounterRecord` looked up `HILL_RADII_AU[body]` directly** while its callers gate on `.get(body, 0.0)`. Safe today because the gate never lets a body without a Hill radius through, but a bare `KeyError` from inside a record builder is a poor way to find that out. It now states the invariant and raises with an explanation.
- **The `"encounters"` JSON key kept its name but lost its guarantee.** It used to hold at most one entry per body; it can now hold several, ordered by time. A downstream reader that assumed one-per-body breaks silently. Noted in the results file and next to the JSON key.
- **`cloneEncounterSummary()` extracted** from `__main__`, so the clone aggregation can be tested directly.
- Comments added to the rest of `hermiteClosestApproach` (the root filter and the candidate comparison).

### `6881186` — tests, 32 → 36

- **A passage outside `n_hill`** must still refine the recorded minimum but must **not** be listed. This is the case that separates the refinement from the encounter detection, and it fails if either half is removed.
- **The Sun** has no Hill radius, so no distance can fall below its threshold and `_encounterRecord` refuses it outright. Both halves are asserted, so the gate and the record builder cannot drift apart.
- **`cloneEncounterSummary`** counts clones that met a body separately from passages made, which is what keeps the reported percentage of the ensemble meaningful when one clone meets the same body twice.

Mutation-checked: reverting the within-step refinement, the Hill-radius gate, or the clone de-duplication fails 5 tests between them.

### One thing I could not exercise

No trajectory pickle on this machine has `traj.uncertainties`, so `--mc` produces no clones and the end-to-end Monte Carlo encounter reporting never ran. `cloneEncounterSummary` is covered by unit tests and a direct render check of the summary lines, but the full CLI path through clones is unverified here.

---

## Merge order for #98, #99 and #100 — verified, not guessed

All three are open, all three branch from `d93fb98`, all three change `REBOUND.py`. I test-merged them for real and ran the suite on the result.

**Recommended order: #98 → #99 → #100.** #98 is the version-compatibility foundation with the smallest semantic surface; #100 is the largest and has to adapt to both.

**Textual conflicts** — one hunk per step, all trivial:

| Step | Conflict | Resolution |
|---|---|---|
| +#99 onto #98 | `heartbeat_ref = _setHeartbeat(sim, heartbeat)` vs this PR's refinement block ending `sim.heartbeat = heartbeat` | Keep this PR's block, end it with #98's `_setHeartbeat` line |
| +#100 | `_checkReboundxAttached(sim)` / `sim.dt = …` | Keep `_checkReboundxAttached(sim)` and **#100's** `sim.dt = 0.001 if integrate_forward else -0.001`. Keeping both `sim.dt` lines *looks* fine and even runs correctly by accident, since the second assignment wins — but it leaves a dead line |
| +#100 | `cloneEncounterSummary` vs `whfastEncounterWarning` (both new, both inserted after `encountersFromMinDistances`) | Keep both |
| +#100 | `encounters = sorted(nominal_diag.get("encounters", []), …)` vs #100's integrator/WHFast/MEGNO block | Keep #100's block but **drop its** `encounters = encountersFromMinDistances(...)` line. Naive "keep both" leaves this PR's list silently overwritten a few lines later |
| +#100 (tests) | Adjacent new sections in `test_REBOUND_numerics.py` | Keep both |

**Semantic follow-ups no merge tool catches** — five, of which two are not in any PR description:

1. **This PR's encounter times need `time_sign`.** `_encounterRecord(b, d, t/(2*np.pi)*365.25)` must become `time_sign*t/(2*np.pi)*365.25`, matching `min_time_days`. Otherwise a backward TRACE run under #100 reports positive encounter times. *(Flagged in #100's description.)*
2. **`computeMegno` must use `_setHeartbeat`.** #100 assigns `sim.heartbeat` directly, which fails on REBOUND 5.1.1 after #98. *(Flagged in #100's description.)*
3. **`whfastEncounterWarning` must read `diag["encounters"]`** instead of calling `encountersFromMinDistances`. **New** — it is in no description, because that function was added during the #100 review. As written it sees at most one encounter per body and under-counts repeats.
4. **The `whfastEncounterWarning` test fixtures must be updated alongside item 3.** **New.** They build synthetic diagnostics in the old `min_dist_au`/`min_time_days` shape; with item 3 applied they need an `"encounters"` list. Without this the merged tree has 2 failures.
5. **#98's lunar closest-approach expectation** moves from −0.0426370612 d (the step-end value) to the dense-reference value. *(Flagged in this PR's description.)* I could not confirm the new number: that test skips with "Example trajectory not found" on my machine.

**Result:** with all five applied, the merged #98+#99+#100 tree gives **75 passed, 1 skipped** (the skip is item 5's test).

Caveat worth stating plainly: I have reviewed #99 and #100 in depth but **not #98**. Its `_setHeartbeat` and `_checkReboundxAttached` appear in the resolutions above on the strength of its description and a read of the merged result, not a review of the PR itself. That review is next.

PR #100, opened 2026-09-18
REBOUND: selectable integrator (IAS15/WHFast/TRACE) and an optional MEGNO chaos run

AstroEloy:
## Summary

Two new command-line options for `wmpl/Rebound/REBOUND.py`, also available as `reboundSimulate` arguments:

- **`--integrator {ias15,whfast,trace}`** (default `ias15`) with **`--dt`** (days; default 0.5) for the fixed-step integrators.
- **`--compute_megno`**: after the requested integration, integrate the nominal solution again with Newtonian gravity only and report the evolution of its MEGNO chaos indicator and whether it converges to 2.

## Integrator choice

The reference is IAS15 on the example trajectory (`20191023_091225`, integrated backward):

| Span | Integrator (dt = 0.5 d) | Relative difference in final a | Wall time |
|---|---|---|---|
| 10 yr | WHFast | 1.5e-6 | 0.18 s (IAS15 0.71 s) |
| 10 yr | TRACE | 2.0e-7 | 0.23 s |
| 100 yr | WHFast | 1.2e-8 | 1.19 s (IAS15 5.05 s) |
| 100 yr | TRACE | 7.5e-9 | 1.68 s |

Getting there required solving three problems:

1. **The departure from the Earth's surface.** The object starts on the Earth, which is a deep close encounter:
   - WHFast cannot resolve it: the final a was off by 13 % to 600 % even with dt from 0.002 to 0.2 d.
   - TRACE resolves the encounter itself, but it applies REBOUNDx's Earth J2/J4 (`gravitational_harmonics`) without sub-stepping, which leaves the orbit about 1–2 % off in a. Without J2 it matched IAS15 to 2e-7.
   - **Fix:** both start with IAS15 and take over at the first output after the object has left the Earth's neighbourhood. This uses the existing departure gate of 3 Earth Hill radii, beyond which the harmonics are negligible. The report states when the handover happened.
2. **TRACE with a negative timestep.** TRACE mishandles close encounters when the timestep is negative. This is a REBOUND bug, present in 4.6.0 and 5.1.1: a minimal Sun–planet–particle flyby matches IAS15 to 7.7e-8 forward but comes out 10 % off in a backward.
   - **Fix:** a backward TRACE run is integrated forward in time with every velocity reversed. This is exact for gravity, `gr_full` (the EIH terms are quadratic in the velocities) and J2/J4.
   - **Radiation forces:** Poynting–Robertson drag is odd in the velocity, so reversed time needs a(r, −v). In REBOUNDx's `radiation_forces`, a = β GM/r² [(1 − ṙ/c) r̂ − v/c]; the drag is entirely in 1/c, so a(r, −v) is the same force with c → −c, which is set on the radiation force in reversed runs. With β = 0.01 and 0.1 over 10 and 100 years, backward TRACE matches backward IAS15 within 2e-6 to 1.3e-5 in a, while radiation itself changes a by 25–190 %. Without the sign change the error is 1e-4 to 1e-3.
3. **TRACE availability.** TRACE only exists from REBOUND 4.4.0, so an older REBOUND gets a clear error instead of `Integrator not found`.

Two further behaviours:
- **WHFast and close encounters:** WHFast does not resolve close encounters. Through flybys inside 0.1 R_Hill of Venus, Earth, Mars and Jupiter, WHFast's error in a was 0.3 % to 200 %, while TRACE stayed within 4e-6; both were fine at about 1 R_Hill. The report therefore warns whenever an encounter under 3 R_Hill happened while WHFast was integrating, for the nominal or any clone.
- **Output times:** the fixed step is adjusted in each output interval to divide it exactly, so the integrator never has to cut a step short to land on an output time.

## MEGNO

`computeMegno` integrates the nominal solution over the requested span with Newtonian gravity only and always with IAS15. IAS15 handles the departure and supports variational equations. TRACE does not support them: REBOUND 4.6.0 aborts the Python process, and 5.1.1 silently returns 0.

- **Only the object's deviation is followed.** A first-order variational test particle is added for the object alone. REBOUND's `init_megno()` perturbs every body, and the linearly growing deviations of the Moon and the planets then hide the object's exponential growth for centuries. For example, a Jupiter crosser with a 64-year Lyapunov time still read <Y> = 2.01 after 300 years with `init_megno()`, but 4.5 with the object alone.
- **How it is computed.** <Y> is accumulated from |δ| at each step, using d ln|δ|/dt = δ·δ̇/|δ|². This uses public API only (`add_variation(testparticle=…)`) and is valid for backward runs. On clean systems (Kepler orbits with e = 0.1 and e = 0.947, and Sun + Jupiter) it matches REBOUND's own MEGNO.
- **Verdict** (`classifyMegno`), based on the final <Y> and its mean over the last quarter of the run:
  - **regular:** both within 0.1 of 2.
  - **chaotic:** both above 2.5; a Lyapunov time is estimated from <Y> ≈ (λ/2) t.
  - **periodic:** both below 1. The deviation stays bounded, e.g. libration in a resonance: regular, but not converging to 2.
  - **not converged:** anything else.
  - **No verdict:** for unbound orbits, or for runs shorter than 20 orbital periods.
- **Example result.** The example Halley-type meteoroid comes out *periodic* (<Y> ≈ 0.5 after 1000 years). A shadow particle confirms it: with the giant planets its separation stays bounded while a librates between 10.67 and 10.98 AU, whereas without them the separation grows steadily.
- **Outputs.** The series and the verdict go to the terminal, the results file and the JSON; a separate `rebound_megno.png` shows the evolution.

## Tests

New `test_REBOUND_integrators.py`, skipped without rebound/reboundx; the TRACE cases are also skipped below 4.4:
- WHFast and TRACE against IAS15 for a departure from the Earth's surface, in both directions.
- TRACE backward with radiation forces (β = 0.01 and 0.1, 10 years) against IAS15.
- MEGNO converging to 2 for a Kepler orbit, forward and backward.
- A Jupiter crosser flagged as chaotic.

`test_REBOUND_numerics.py` gains the MEGNO verdict numerics. Each new test was checked to fail when its fix is removed (no time reversal, no IAS15 departure phase, no c sign change for radiation, whole-system MEGNO). All 39 Rebound tests pass on rebound 4.6.0; on 4.3.0, 35 pass and the 4 TRACE tests are skipped.

## Interaction with #98 and #99

All three PRs change `REBOUND.py`. I merged them together locally; the textual conflicts are small:

- **Resolutions:**
  - Keep `_checkReboundxAttached(sim)` next to the new `sim.dt` line.
  - Keep #99's chronological encounter list next to the new integrator and MEGNO block.
  - Keep both new sections in `test_REBOUND_numerics.py`.
- **Two changes that no merge tool will catch:**
  1. #99's `"encounters"` diagnostic must convert its times with `time_sign`, like `min_time_days`. Otherwise encounters in a backward TRACE run get positive times.
  2. `computeMegno` must set its heartbeat with #98's `_setHeartbeat` (and keep the returned reference). The plain `sim.heartbeat = …` fails on REBOUND 5.1.1.

With these, all 59 Rebound tests pass on rebound 5.1.1 and 4.6.0; on 4.3.0, 55 pass and the 4 TRACE tests are skipped. With #99, the encounter minima are also refined within the 0.5-day fixed steps: a Mercury flyby gives the same minimum with all three integrators to 1e-5.

## Notes

- **REBOUNDx warning:** REBOUNDx 5 warns that `gr_full` is velocity-dependent when used with WHFast. Its measured effect is below 1e-8 in a over 100 years, and the `--integrator` help says so.
- **TRACE bug not reported upstream:** the negative-timestep TRACE bug should probably be reported to REBOUND; this PR only works around it.


dvida:
## Review pass: 3 commits pushed to this branch

Reviewed for correctness, style, performance, comment density, tests and documentation. **The physics is sound** — all 39 tests passed as submitted, on rebound 4.5.0 / reboundx 4.6.0, and I found no bug in the integrator hand-over, the reversed-time TRACE workaround or the MEGNO accumulation. The three commits below fix smaller issues, close test gaps, and add the manual this module never had.

Verified end to end on three real trajectory pickles (TRACE and WHFast, with `--compute_megno`), which between them exercised the `unbound`, `too_short` and `not_converged` verdicts and fired the WHFast close-encounter warning on a real Jupiter flyby. Integrator results are bit-identical before and after the refactor.

### `fa03965` — correctness and comments

- **`--dt` was not validated.** `--dt 0` or a negative value fell through to `n_sub = max(1, ...)` and collapsed to a single step spanning an entire output interval — silently wrong rather than merely inaccurate. Now refused in both the CLI and `_integrateParticles`. The `task.get("dt_days") or ...` fallback is also replaced with an explicit `None` check, since `or` swallows `0.0`.
- **`_grForceName(integrator)` removed.** It took an argument it ignored, always returned `"gr_full"`, and had one call site. The original comment it replaced is restored.
- **`checkIntegratorAvailable`** appended "TRACE was added in REBOUND 4.4.0" for any unavailable integrator, including `whfast`. Now conditional.
- **`compute_megno` ran even with `return_diagnostics=False`**, doing the full extra integration and then discarding it, since the diagnostics are the only way it is returned. Now skipped with a warning.
- **`rebound_megno.png` plotted days.** A run only gets a verdict past 20 orbital periods, so the span is always 10^4–10^6 days and the axis was unreadable. Years now; `times_days` in the JSON is unchanged.
- **`whfastEncounterWarning()` extracted** from the `__main__` block so it can be tested directly.
- **Comments.** `classifyMegno` had zero inline comments across ~50 lines; `computeMegno` and `_megnoReportLines` were thin. Brought to the density used elsewhere in wmpl (roughly one per 6–9 lines, whole-line above the code, `#   ` hanging continuation). Also recorded *why* WHFast has to stay in its default safe mode: `move_to_com`/`move_to_hel` modify the particles at every output and the heartbeat reads them at every internal step, so turning it off would be faster but would silently corrupt both.

### `e56bd33` — tests, 39 → 52

New cases: non-positive `--dt` rejected and an omitted one falling back to the default; the 4.4.0 hint appearing only for TRACE; `classifyMegno` with fewer than 8 points and with none at all; `_megnoReportLines` per verdict status including the empty series; and `whfastEncounterWarning` counting encounters after the hand-over, for clones as well as the nominal solution, while ignoring earlier ones and particles that never handed over.

Each new test was checked to fail when its fix is reverted, matching the standard set in the PR description.

Also: `reb.rb` instead of `__import__("rebound")`, a missing docstring on `_finalOrbit`, assert messages on the non-obvious checks, a corrected launch speed in `_departureTask`'s docstring (it is `sqrt(20^2 + 11.1^2)` km/s, not 20), and a note in the module docstring on why this file skips rather than mocking the imports like the other Rebound tests.

### `de87f85` — `wmpl/Docs/`

The library had no documentation outside the docstrings: the root README has no relative links at all, never mentions `wmpl.Rebound`, and the one complete manual in the tree (`wmpl/Dynesty/README.md`) is linked from nowhere.

- **`wmpl/Docs/REBOUND.md`** — installing rebound/reboundx, the pickle and ephemeris it reads, every CLI flag, choosing between IAS15/WHFast/TRACE and how to check a fixed-step run was safe, radiation forces, close encounters and impacts, Monte Carlo clones, MEGNO and its verdicts, all four output files including the JSON layout, seven worked examples, the library API, and troubleshooting.
- **`wmpl/Docs/Trajectory.md`** — producing the trajectory pickle the other tools consume.
- **`wmpl/Docs/README.md`** — index, also linking the existing Dynesty manual.
- Root README gains a `## Documentation` section and a REBOUND invocation in `## Usage`. All relative links and TOC anchors verified.

Formatting follows what `wmpl/Dynesty/README.md` already established: anchor table of contents, `| Flag | Description |` tables, `---` rules, untagged shell fences.

---

### One measurement worth flagging

The description reports WHFast and TRACE landing within ~1e-8 of IAS15 in the final `a` at `--dt 0.5`. On a synthetic Earth-departure task with a lunar encounter, over the same 100-year span, I measured **2.0e-4 (WHFast) and 6.5e-5 (TRACE)** — four orders of magnitude looser. The 1e-8 figure looks specific to the example trajectory rather than general.

Nothing in the code needs to change for this: the code comments and the `--integrator` help already point at the same caveat, and the manual now states the accuracy as case-dependent and recommends comparing one shorter run against IAS15 before trusting a long fixed-step result. It may be worth softening the claim in the PR description too.

### Two description-only notes

- ~~The **"Interaction with #98 and #99"** section is stale. This branch is already based on the current master tip (`d93fb98`), master uses a plain `sim.heartbeat = ...` with no `_setHeartbeat`, and the `"encounters"` times already carry `time_sign`. Nothing left to resolve.~~

  **Correction (edited):** I had this backwards. #98 and #99 are both still **open**, so the "Interaction" section is live and correct, not stale. The absence of `_setHeartbeat` from master is evidence that #98 has *not* merged, which is the opposite of what I wrote. Apologies for the noise.

  One item to add to that section, which is in neither description because it came out of this review: **`whfastEncounterWarning` (added above) calls `encountersFromMinDistances`**, so once #99 lands it should read `diag["encounters"]` instead — otherwise it sees at most one encounter per body and under-counts repeated passages.
- The TRACE negative-timestep bug is still only worked around here, not reported upstream — worth doing, as the description itself notes.

dvida:
Follow-up: one more commit on this branch, `864bcfb`.

`wmpl/Docs/REBOUND.md` described the pre-#99 behaviour — encounter minima sampled at the step ends and called "exact", and one entry per body. #99 refines each minimum within the step and lists every passage, so the manual would have shipped wrong as soon as both land.

Updated: why both the per-timestep tracking and the within-step refinement are needed, that the list is ordered by time and a body can repeat, that only the deepest passage per body is labelled on the plot, the `clone_encounters` JSON key, the `count` vs `n_encounters` distinction under `clone_outcomes`, and three new entries in the library API table.

This documents behaviour that arrives with #99, so it is only correct once both PRs are merged.

I have also posted a verified merge-order analysis for #98, #99 and #100 on [#99](https://github.com/wmpg/WesternMeteorPyLib/pull/99#issuecomment-5745002819). Two items in it concern this PR and are in no description, because they came out of these reviews:

- `whfastEncounterWarning` (added here) must read `diag["encounters"]` after #99 lands, rather than calling `encountersFromMinDistances` — otherwise it sees at most one encounter per body and under-counts repeated passages.
- The `whfastEncounterWarning` test fixtures need an `"encounters"` list once that change is made; without it the merged tree has 2 failing tests.

With those and the three already-known follow-ups applied, the merged #98+#99+#100 tree gives 75 passed, 1 skipped.

PR #102, opened 2026-09-19
List close approaches to the Sun with the other close encounters

AstroEloy:
## Summary

The Sun has no Hill radius, so `REBOUND.py` tracked its distance but never listed it as a close encounter, and it could not be hit either. This adds both, with a small change.

- **Close approaches.** A perihelion passage closer than `SUN_ENCOUNTER_AU = 0.1` AU goes into the same encounter list as the planets, in time order. The Sun gets a fixed distance threshold, not a fake Hill radius, so its entries have `hill_radius_au` and `n_hill` set to `None`. The terminal and report lines show the distance in solar radii instead of R_Hill, and the headers state both rules:
  ```
  Close encounters (< 3 Hill radii, Sun < 0.10 AU), in the order they happened:
  ```
- **Impacts.** `SUN_RADIUS_AU` (695,700 km) is added to the default `body_radii`, so an orbit that dives into the Sun ends in an impact.
- The WHFast warning counts Sun passages too, since WHFast degrades at small perihelia.
- The JSON gets `sun_encounter_au` next to `n_hill_threshold`. Readers of `"encounters"` must allow `n_hill` to be `None`, and the example in `REBOUND.md` is updated for this.

## Tests

- Replaced the test asserting that the Sun is never an encounter with two new tests: a Sun passage at 0.05 AU is listed and one at 0.15 AU is not, and a perihelion inside the Sun is an impact.
- 84 passed on rebound 5.1.1 / reboundx 5.1.0.
- Mutation-checked: removing the Sun gate, the Sun radius, or the record builder's acceptance of the Sun each fails one of the new tests.
- Ran the CLI on the example trajectory (forward, with a temporarily wider threshold) and checked the terminal, report file and JSON output.

## Notes

- The threshold is a module constant, like `n_hill`.
- A stream with perihelion below 0.1 AU gets one Sun line per perihelion passage. The δ Aquariids (q ≈ 0.07 AU) would get about 20 over 100 years; this is estimated from the orbit, not run.


dvida:
## Review

The change is correct and stays as written. I checked it independently of the numbers in the description (scratch harness on rebound 4.5.0 / reboundx 4.6.0, then the real checkout), found one consistency gap, and pushed three commits on top: `f848616`, `758bfa2`, `166fdc0`.

### Verified

| Check | Result |
| --- | --- |
| Test suite | 83 passed + 1 skipped in the scratch harness (= your 84, the skip is the integration-pickle test, which only runs in a real checkout); **90 passed** in the real checkout with the new tests |
| Mutation: Sun gate set to 0 | fails `testACloseSunPassageIsAnEncounter`, as claimed |
| Hermite refinement at perihelion (single output over the span vs 3001 outputs) | rel. diff 1.2e-6 at q = 0.05 and 0.02 AU, 9.6e-5 at q = 0.005 AU (≈1 R☉), times within 0.03 min |
| Impact edge | q = 0.5 / 0.9 / 0.99 R☉ → impact recorded at 0.99 R☉, no encounter entry; q = 1.01 / 1.10 R☉ → no impact, one Sun encounter; `body_radii={"Sun": 0}` disables it |
| Backward run | a Sun passage 13.37 d into the past is listed at t = −13.37 d, equal to `min_time_days["Sun"]` |
| `cloneEncounterSummary`, `whfastEncounterWarning` | both handle a Sun record |
| Merge | 0 conflicts against master and against #97 |

Your two tests only assert `rel=1e-2`; the intent (listed / not listed) is right and the planetary tests in the same file assert 1e-6, so I added the accuracy test rather than tightening yours.

### Found

**`encountersFromMinDistances` did not know the Sun.** For the same diagnostics it returned `[]` while the per-particle `encounters` list held the Sun passage, although its docstring calls it the way to rebuild that list from the closest-approach dictionaries. It iterated `HILL_RADII_AU` only and sorted on `n_hill`, which would have crashed on a `None`. Fixed in `f848616`: it now applies the same gates as the heartbeat through a shared `_encounterThreshold(body, n_hill)` helper and sorts by closeness in units of each body's own threshold. `detectCloseEncounters` (the legacy scan over sampled output) now says explicitly that it does not list the Sun.

**The JSON did not record the gates for a nominal-only run.** `n_hill_threshold` and `sun_encounter_au` sat only under `clone_outcomes`, which is `null` without `--mc`, and a Sun record's `n_hill = None` cannot be inverted to recover the gate. Both are now also in `run`; the `clone_outcomes` copies are kept.

**The WHFast claim is understated.** A δ-Aquariid-like orbit (a = 2.6 AU, q = 0.07 AU, i = 27°), 100 yr forward, Sun + 5 planets on circular orbits:

| Integrator | dt | final a vs IAS15 |
| --- | --- | --- |
| IAS15 | adaptive | reference, a = 2.5587 AU |
| WHFast | 0.5 d (default) | **ejected, a = −8.9 AU** |
| WHFast | 0.1 d | 1.7e-4 |
| TRACE | 0.5 d (default) | 7.2e-4 |
| TRACE | 0.1 d | 2.0e-9 |

So counting Sun passages in the warning is right, but the warning text quoted only the planetary-flyby figures and the manual's WHFast section did not mention perihelia. Both now do, with the table and the advice to use IAS15 or TRACE with `--dt 0.1` inside ~0.1 AU.

**Report volume, measured rather than estimated.** The same orbit gives 24 Sun lines against 6 planetary flybys per 100 yr, so the manual's 1000-yr TRACE example would print ~240 Sun lines. Denis chose to collapse them: when the Sun has more than three passages, the terminal and the text report print one line in place of them, where the first passage was, so the time order of the rest of the list is kept:

```
Sun      24 perihelion passages < 0.10 AU: q = 0.0441-0.0645 AU (9.5-13.9 R_Sun), t = +3.1 to +99.6 d
```

Three or fewer are listed one by one as before, planets are never collapsed, and the JSON keeps every passage. The rule is in `sunPassageSummary()` / `formatSunPassageSummary()` so both outputs share it and it is tested on plain dicts (`758bfa2`). I exercised the real terminal and report loops with a collapsed list and the CLI end to end on a real trajectory pickle (no Sun passage there: header, `none detected`, `run.sun_encounter_au` all present). There is no low-perihelion pickle on this machine, so the collapse through the full CLI rests on the helper's unit test plus the loop exercise.

**Why 0.1 AU** was not stated anywhere. It is a good value, and the code and manual now say why: Na is thermally desorbed at q < 0.1 AU (Kasuga et al. 2006, A&A 453, L17; the Na-free meteoroids of Borovička et al. 2005), asteroids disrupt at q ≈ 0.076 AU (Granvik et al. 2016, Nature 530, 303), and it lies inside the sunskirter zone, q < 33 R☉ = 0.153 AU (Jones et al. 2018, SSRv 214, 20). The manual also notes that a Sun passage is a thermal event rather than a gravitational perturbation. `SUN_RADIUS_AU` is labelled as the IAU 2015 nominal value.

### Tests added (all mutation-checked: each new behaviour reverted fails exactly its own test)

- `testEncountersFromMinDistancesIncludesTheSun` (numerics): listed with `n_hill None`, sorted by its own threshold, dropped outside 0.1 AU whatever `n_hill` is
- `testSunMinimumIsNotLimitedByTheIntegratorStep`: single output vs dense at q = 0.02 AU, `rel=1e-5`
- `testABackwardSunPassageIsListedAtNegativeTime`
- `testAPerihelionInsideTheSunIsAnImpact` parametrized over 0.5 and 0.99 R☉ (impact at the surface, no encounter entry), plus `testAPerihelionJustAboveTheSurfaceIsAnEncounterNotAnImpact` at 1.01 R☉
- `testSunPassageSummaryCollapsesOnlyRepeatedPassages`

After the refactor the stream run gives the same 24 passages and the same final a to 16 digits. Ready to merge from my side.

